PR #11, opened 2016-07-15
[WIP] Deploy

lmcglohon:


Capfile
```diff
@@ -21,10 +21,10 @@ require 'capistrano/deploy'
 # require 'capistrano/rails/assets'
 # require 'capistrano/rails/migrations'
 
-require 'capistrano/bundler'
-require 'capistrano/rails'
+#require 'capistrano/bundler'
+#require 'capistrano/rails'
 require 'dlss/capistrano'
-require 'capistrano/rvm'
+#require 'capistrano/rvm'
```
drh-stanford: The `Gemfile` will likely need these in the `:deployment` section -- I don't think `dlss-capistrano` requires all of them as dependencies.

config/deploy.rb
```diff
@@ -1,11 +1,11 @@
 # config valid only for Capistrano 3.1
-lock '3.2.1'
+#lock '3.2.1'
```
drh-stanford: I would prefer to delete these comments -- this locking business is no longer needed and was historical from the 2.x -> 3.x Capistrano upgrade as I recall.

PR #12, opened 2016-07-15
remove unneeded mysql gem

drh-stanford:
This PR removes the `mysql` gem from `Gemfile`. `mysql2` should be sufficient.


mejackreed:
Its unclear to me, why this project needs a database at all. :shrug:

PR #21, opened 2016-07-18
Remove db, and active_record.

### LINKED ISSUE #14: Check in a database.yml
Similar to #13 .. Kill the setup task if we can.

### LINKED ISSUE #20: remove unused database references
There's no ActiveModel classes and an empty `db/schema.rb` -- can we delete the `database.yml` and all references to a database in the configuration?

mejackreed:
Fixes #20. Fixes #14 

Would like some eyes on this to see if we might need active_record in the future. Potentially we may need it if we implement some delayed jobs type thing.


drh-stanford:
Missed a few db tasks here: https://github.com/sul-dlss/purl-fetcher/blob/remove-db/lib/tasks/ci.rake#L10-L15

drh-stanford:
Also, would need to submit PRs/Issues to puppet repo for purl-fetcher...

mejackreed:
Thanks for the feedback. Updated PR to remove those referenced rake tasks. In regards to Puppet.. I'm not exactly sure what needs to be done there to remove everything.

@eefahy Is that just as simple as removing the `profile::mysql` from the hieradata node?

drh-stanford:
migrated the puppet issue to https://github.com/sul-dlss/puppet/issues/686

Issue #30, opened 2016-07-18
deploy to stage


drh-stanford:
including all configuration

PR #34, opened 2016-07-19
rspec is optional on deployed machines

drh-stanford:
This PR changes the Gemfile such that `rspec` is optional. Otherwise it breaks the `Rakefile` on deployed machines... is connected to #30


eefahy:
Looks good to me but I noticed @lmcglohon changed this in `928a798`

mejackreed:
👍

drh-stanford:
It'll cause failed rake tasks on the deployed machines otherwise... like so:

```
rake aborted!
LoadError: cannot load such file -- rspec/core/rake_task
```

PR #45, opened 2016-07-20
use solr.yml and secrets.yml as-is for testing/development

### LINKED ISSUE #13: Check in a secrets.yml
Rather than local devs creating one, just check one in. We can then set it as a `linked_file` in our deploy and set via shared configs.

https://github.com/sul-dlss/earthworks/blob/master/config/deploy.rb#L27

drh-stanford:
This PR fixes #13. It removes the need to manage the `solr.yml` and `secrets.yml` in development/test environments.

The deploy already has `solr.yml` and `secrets.yml` in the `linked_files` setup so it *should* overwrite them on deploy. Need to test that...


eefahy:
👍  - However I would like to verify that the proper files from `shared_configs` is symlinked in properly on deploy

lmcglohon:
@eefahy Is this ready to merge?

drh-stanford:
I've deployed this on dev and it symlinks ok

```
$ cd ~/purl-fetcher/current/
$ ls -l config/{secrets,solr}.yml
lrwxrwxrwx 1 lyberadmin lyberteam 58 Jul 25 10:17 config/secrets.yml -> /opt/app/lyberadmin/purl-fetcher/shared/config/secrets.yml
lrwxrwxrwx 1 lyberadmin lyberteam 55 Jul 25 10:17 config/solr.yml -> /opt/app/lyberadmin/purl-fetcher/shared/config/solr.yml
```

PR #50, opened 2016-07-21
cleanup ci.rake tasks

peetucket:
remove unnecessary tasks; update readme; add default rake task for running tests; add rake task for rebuild cassettes

### DIFF
--- a/lib/tasks/ci.rake
+++ b/lib/tasks/ci.rake
@@ -1,16 +1,24 @@
 require 'jettywrapper' unless Rails.env.production? || Rails.env.development?
 require 'rest_client'
 
-desc 'Run continuous integration suite (assuming jetty is not yet started)'
+desc 'Run continuous integration suite (tests, coverage, rubocop, docs)'
 task :ci do
-  Rake::Task['rubocop'].invoke
+   Rake::Task['rspec'].invoke
+   Rake::Task['rubocop'].invoke
+end
+
+desc 'Rebuild vcr cassettes and run tests (assuming jetty is not yet started)'
+task :rebuild_cassettes do
   unless Rails.env.test?
-    system('bundle exec rake ci RAILS_ENV=test')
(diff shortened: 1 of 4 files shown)


drh-stanford:
I rebased on master to get #53 and pushed up. Passes Travis now.

lib/tasks/ci.rake
```diff
@@ -1,16 +1,29 @@
...
-desc 'Run continuous integration suite (assuming jetty is not yet started)'
+desc 'Run continuous integration suite (tests, coverage, rubocop, docs)'
 task :ci do
-  Rake::Task['rubocop'].invoke
   unless Rails.env.test?
     system('bundle exec rake ci RAILS_ENV=test')
   else
+    system('bundle exec rspec spec --color')
+    Rake::Task['rubocop'].invoke
+    system('yard doc')
```
drh-stanford: can we omit building the documentation in the ci build? i don't think the output docs are used by anything.

peetucket: sure.. just did it for completeness but easy enough to omit

lib/tasks/ci.rake
```diff
@@ -1,16 +1,29 @@
...
 require 'rest_client'
 
-desc 'Run continuous integration suite (assuming jetty is not yet started)'
+desc 'Run continuous integration suite (tests, coverage, rubocop, docs)'
 task :ci do
-  Rake::Task['rubocop'].invoke
   unless Rails.env.test?
     system('bundle exec rake ci RAILS_ENV=test')
   else
+    system('bundle exec rspec spec --color')
```
drh-stanford: we already have a `.rspec` file for configuration so this can just be `bundle exec rspec` i think

peetucket: and possibly just Rake::Task['rspec']

peetucket: i'm not sure you need to explicitly run rails_env=test either

Issue #46, opened 2016-07-20
Improve (add) Logging


blalbrit:
* [x] log what druid & what action is being taken on the druid for indexer
* [x] log request & if request was fulfilled for fetcher --  who (external user vs dispatcher auto-mods), what (GET, parms), when, (most already in Apache log for incoming)
* [ ] request/response information viewable together, e.g., in squash or honeybadger equivalent)
* [x] consolidate log files

Existing:
* rails environment logging
* delete rake task log
* indexing rake task log

drh-stanford:
The apache logs will have the who/what/when for the incoming requests so we don't need to log those in the application itself.

drh-stanford:
I've spawned the final logging requirement into https://github.com/sul-dlss/purl-fetcher/issues/58 as I'm not clear on the actionable work for it.

PR #51, opened 2016-07-22
add logging for each item processed and sent to solr during indexing

drh-stanford:
This PR is connected to #46. It adds a minimal set of logging for indexing items (adds and deletes).

### DIFF
--- a/lib/indexer.rb
+++ b/lib/indexer.rb
@@ -258,8 +258,13 @@ module Indexer
     solr = establish_solr_connection
     response = {}
     begin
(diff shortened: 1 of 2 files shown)


drh-stanford:
Looks like there might be a race condition in the specs -- the push succeeded but the pr failed Travis. Need to investigate...

drh-stanford:
Depends on #52 to be fixed...

drh-stanford:
Rebased and ready for review now.

Issue #22, opened 2016-07-18
Review field names in the canned data for local purl-fetcher versus field names in sul-solr indexes


lmcglohon:
for purl-dev, -stage, and -prod.  Fix them the match expected field names.

lmcglohon:
For example, a record in purl-dev, -stage, and -prod looks like this:
{
        "title_tsi":"Santa Rosa Races #2, 1954.",
        "public_dc_title_t":["Santa Rosa Races #2, 1954."],
        "id":"druid:bb298yx8728",
        "true_releases_ssim":["revs_stage"],
        "false_releases_ssim":["Atago",
          "Robot_testing_feb_5_2015"],
        "identityMetadata_objectType_t":["item"],
        "identityMetadata_objectType_tsim":["item"],
        "is_member_of_collection_s":["druid:kh678dr8608",
          "druid:nt028fd5773"],
        "is_member_of_collection_ssim":["druid:kh678dr8608",
          "druid:nt028fd5773"],
        "indexed_dtsi":"2015-07-03T01:00:56Z",
        "published_dt":"2015-07-03T01:00:56Z",
        "_version_":1505634819279683584,
        "timestamp":"2015-07-03T01:00:57.047Z"}

A record in the local solr index created by the fixture data looks like this:
  "id":"druid:bb001zc5754",
        "title_tesim":["French Grand Prix and 12 Hour Rheims: 1954"],
        "objectType_ssim":["item"],
        "published_dttsim":["2014-05-05T05:04:13Z",
          "2014-06-06T05:04:13Z",
          "2014-06-06T05:06:06Z"],
        "is_governed_by_ssim":["info:fedora/druid:qv648vd4392"],
        "is_member_of_ssim":["info:fedora/druid:nt028fd5773",
          "info:fedora/druid:wy149zp6932"],
        "catkey_id_ssim":["3051728"],
        "_version_":1539946494904238080,
        "timestamp":"2016-07-15T18:30:20.579Z"}

lmcglohon:
fixed by https://github.com/sul-dlss/purl-fetcher/pull/55

PR #55, opened 2016-07-22
Clean up and remove need for solr_terms.yml

lmcglohon:
PR is connected to #22 
Rename Fedora types to object types
Removed default of fedora_prefix, druid_prefix, and the three object types
Removed id, type, and catkey fields
Removed title_field and title_field_alt variables
Updated field name for is_member_of_ssim to is_member_of_collection_ssim, removed controller_type and collection field
Removed last_changed field and remove solr_terms.yml


lmcglohon:
@mejackreed In order to remove solr_terms.yml, I had to create a class variable called @@object_types. Do you have a better suggestion?

mejackreed:
Why does it need to be a class variable? Is `object_types` a constant? Maybe then make it a constant?

I think the problem for me with a class variable here, is that this is a module, not developed as a class. If its not mutated I would opt for a constant, if it is muted potentially something else.

lmcglohon:
@mejackreed The collection controller needs access to it as well. If I make it a constant, I'm not sure where I should put it...

mejackreed:
I would say a constant at the top of the module then.. 

```ruby
module Fetcher

OBJECT_TYPES = { collection: 'collection', item: 'item', set: 'set'}
```

lib/fetcher.rb
```diff
@@ -18,32 +20,32 @@ def run_solr_query(params, method = 'select')
...
   # (potentially limited by rows or date if specified by the user)
   # @param params [Hash] querystring parameters from user, which could be an empty hash
-  # @param ftype [String] fedora object type, could be :collection
+  # @param ftype [String] object type, could be :collection
   # @return [Hash] solr response
-  # @example find_all_fedora_type(params,:collection)
-  def find_all_fedora_type(params, ftype)
+  # @example find_all_object_type(params,:collection)
+  def find_all_object_type(params, ftype)
     # ftype should be :collection (or other symbol if we added more since this was updated)
```
drh-stanford: might be good to ensure that `OBJECT_TYPES.include?(ftype)` or fail with `ArgumentError`, otherwise the Solr query will fail silently.

lmcglohon: Good idea @drh-stanford   I'll add that in.

PR #68, opened 2016-08-03
complete rebuild of how finding and indexing of files occur

peetucket:
also, remove dor-fetcher code and rewrite most of the tests


lib/purl_parser.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+    @releases
+  end
+
+  # Extract the druid from publicXML identityMetadata.
+  #
+  # @return [String] The druid in the form of druid:pid
+  #
+  def druid
+    @druid ||= public_xml.xpath('//publicObject')[0].attr('id')
```
drh-stanford: this can be an `at_xpath` call rather than `xpath(...)[0]`

app/models/run_log.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  # the last completed run, returned as a model object
+  def self.last_completed_run
+    self.order('ended DESC').find_by('ended IS NOT NULL')
+  end
+
+  # remove all rows with no end time that were started more than 2 days ago (i.e. the job was started and must have died without completing the entry)
+  def self.prune_crashed_rows
+    self.where('updated_at < ?', 2.days.ago).where('ended IS NULL').each(&:destroy)
```
drh-stanford: you may want to make the timeframe be configurable -- is it used anywhere else?

PR #77, opened 2016-08-08
implements simple /collections API

### LINKED ISSUE #2: PURL Fetcher collection call does not return members
PURL Fetcher is getting no results from a membership query for either collections

The correct field names are mostly correct in the solr_terms.yml in github
title_tsi
public_dc_title_t
id
true_releases_ssim
false_releases_ssim
identityMetadata_objectType_t
identityMetadata_objectType_tsim
is_member_of_collection_s
is_member_of_collection_ssim
indexed_dtsi

published_dt

version

timestamp

Purl fetcher does not support return an APO member query

drh-stanford:
This PR fixes #2 and implements a bare bones API for `/collections`. Namely:

* `GET /collections` returns a JSON array of all PURL collections druids
* `GET /collections?rows=0` returns count of all PURL collections
* `GET /collections/druid:aa111bb2222` returns JSON array of all PURL druids in the collection
* `GET /collections/druid:aa111bb2222?rows=0` returns count of all PURLs in the collection

As of 8/8, the design of the full API is still TBD. As such, I resurrected only the most basic level of functionality from PR #68. The current design document lives here: https://consul.stanford.edu/display/GRYPHONDOR/PURL+Fetcher

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -1,0 +1,54 @@
+class CollectionsController < ApplicationController
+  ##
+  # API call to get a full list of all PURL collections
+  #
+  # @param [querystring] Parameters can be specified in the querystring
+  #   * rows = number of results to return (set to 0 to only get count)
+  #
+  #  http://localhost:3000/collections
+  #  http://localhost:3000/collections?rows=0
+  #
+  # @return [Array<String>] list of druids for all PURL collections
+  #
+  #
+  #
+  def index
+    collections = Purl.where(object_type: %w(set collection))
+    if params[:rows] == '0'
+      respond_to do |format|
+        format.json { render json: { size: collections.size}.to_json }
+      end
+    else
+      respond_to do |format|
+        format.json { render json: collections.to_a.map(&:druid).to_json }
+      end
+    end
+  end
+
+  ##
+  # API call to get a list of PURL druids associated with a specific collection
+  #
+  # @param [querystring] Parameters can be specified in the querystring
+  #   * rows = number of results to return (set to 0 to only get count)
+  #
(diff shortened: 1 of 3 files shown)


spec/features/collections_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  it 'fetches all purls for a collection' do
+    get '/collections/druid:oo000oo0001'
+    expect(response).to be_success
+    results = JSON.parse(response.body)
+    expect(results).to be_an(Array)
+    expect(results).to eq ['druid:bb1111cc2222', 'druid:dd1111ee2222']
+  end
+
+  it "fetches all collections with size only" do
```
lmcglohon: Shouldn't this be it "fetches a specified collection with size only" do ?

app/controllers/collections_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  #
+  #  http://localhost:3000/collections/druid:aa111bb2222
+  #  http://localhost:3000/collections/druid:aa111bb2222?rows=0
+  #
+  # @return [Array<String>] list of druids for all PURLs that are members of the given collection
+  #
+  def show
+    druid = params[:id]
+    fail ArgumentError, "Invalid druid: #{druid}" unless druid.present?
+    fail ArgumentError, "Invalid collection: #{druid}" unless Collection.find_by_druid(druid).present?
```
mejackreed: I think Rails will automatically give us this, no?

If not could we just use strong params?

```ruby
def druid_param
  params.require(:druid)
end
```

drh-stanford:
I changed to use `params.require` and I revised all of the spec descriptions.

mejackreed:
👍 thanks!

PR #86, opened 2016-08-09
Add pagination

mejackreed:
Note: This pr changes the default sort to latest_change which is different than requested by @blalbrit but works better with discovery dispatcher.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gem 'stanford-mods'
 gem 'retries'
 gem 'druid-tools'
 gem 'whenever', :require => false
+gem 'kaminari' # for pagination
 
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 gem 'jbuilder', '~> 2.0'
--- a/spec/controllers/docs_controller_spec.rb
+++ b/spec/controllers/docs_controller_spec.rb
@@ -34,7 +34,7 @@ RSpec.describe DocsController do
       end
       it 'page' do
         get :deletes, format: :json, per_page: 1, page: 2
-        expect(assigns(:deletes).first.druid).to eq 'druid:cc1111dd2222'
+        expect(assigns(:deletes).first.druid).to eq 'druid:ff1111gg2222'
         expect(assigns(:deletes).count).to eq 1
       end
     end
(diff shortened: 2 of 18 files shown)


Gemfile
```diff
@@ -7,6 +7,7 @@ gem 'stanford-mods'
 gem 'retries'
 gem 'druid-tools'
 gem 'whenever', :require => false
+gem 'kaminari'
```
drh-stanford: maybe add that we use it for pagination

spec/controllers/docs_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        expect(assigns(:deletes).count).to eq 1
+      end
+      it 'page' do
+        get :deletes, format: :json, per_page: 1, page: 2
+        expect(assigns(:deletes).first.druid).to eq 'druid:ff1111gg2222'
+        expect(assigns(:deletes).count).to eq 1
+      end
+    end
+  end
+end
```
drh-stanford: is there a provision for too large a page size in the input parameters? is that the `max_paginates_per` setting?

spec/features/collections_controller_spec.rb
```diff
@@ -22,7 +22,7 @@
     expect(response).to be_success
     results = JSON.parse(response.body)
     expect(results).to be_an(Array)
-    expect(results).to eq ['druid:bb1111cc2222', 'druid:dd1111ee2222']
+    expect(results).to eq ['druid:dd1111ee2222', 'druid:bb1111cc2222']
```
drh-stanford: these are order-sensitive now, right? i'd add a comment to that effect

drh-stanford:
Looks good to me 👍 . I added a few minor comments

spec/controllers/docs_controller_spec.rb
mejackreed: Yes, we max it out at 10,000

Gemfile
mejackreed: added.

spec/features/collections_controller_spec.rb
mejackreed: Added

mejackreed:
Thanks for the comments @drh-stanford . I've pushed up changes to address them.

PR #95, opened 2016-08-09
move from indexed_at to published_at, also use public xml modified time

### LINKED ISSUE #92: Move indexed_at to published_at
This value should be learned from the publicxml doc modified time from the `public` file. We can go ahead and kill `indexed_at` as Rails gives us `created_at` and `updated_at` for free.

cc @drh-stanford @eefahy

mejackreed:
Closes #92

### DIFF
--- a/lib/purl_parser.rb
+++ b/lib/purl_parser.rb
@@ -1,12 +1,13 @@
 # class used to represent a purl, used to parse information
 class PurlParser
 
-  attr_reader :public_xml # the public XML as a nokogiri doc (will be nil if the public file was not found)
+  attr_reader :path, :public_xml # the public XML as a nokogiri doc (will be nil if the public file was not found)
 
   # Given a path to a directory that contains a public xml file, read in the full XML and set the public_xml accessor as a nokogiri document
   # @param path [String] The path to the directory that contains the public file
   #
   def initialize(path)
+    @path = path
     begin
       @public_xml ||= Nokogiri::XML(File.open(Pathname(path) + 'public'))
     rescue => e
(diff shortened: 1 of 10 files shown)


spec/features/purl_parser_spec.rb
```diff
@@ -51,4 +51,13 @@
...
     end
   end
+  describe '#modified_time' do
+    let(:sample_doc_path) do
+      DruidTools::PurlDruid.new('bb050dj7711', purl_fixture_path).path
+    end
+    subject { described_class.new(sample_doc_path) }
+    it 'gets the file modified_time from file system' do
+      expect(subject.modified_time).to be_an Time
+    end
```
drh-stanford: we should verify that the `modified_time` is coming from the file system. something like this maybe:

```
before_modified = Time.now
FileUtils.touch(sample_doc_path)
expect(subject.modified_time).to be > before_modified
```

drh-stanford:
one comment on the specs, but otherwise 👍

lib/purl_parser.rb
```diff
@@ -1,12 +1,13 @@
 # class used to represent a purl, used to parse information
```
drh-stanford: do we need to `require 'pathname'` here for clarity?

mejackreed: I'm not sure... Its part of the standard library and should be required within a Rails application. I could go either way :) Happy to add it if you think it would be helpful

spec/features/purl_parser_spec.rb
mejackreed: Great comment thanks. added.

lib/purl_parser.rb
drh-stanford: It looks like rails always requires `pathname` but I don't think it does that for other parts of stdlib. I can't find any documentation on this point though. Here's the pathname require: https://github.com/rails/rails/blob/4-2-stable/railties/lib/rails.rb#L3

PR #98, opened 2016-08-10
catch errors on PurlFinder#index_purls

### LINKED ISSUE #99: PurlFinder.index_purls should log rather than raise exceptions
See https://github.com/sul-dlss/purl-fetcher/blob/master/lib/purl_finder.rb#L88-L89

There should be a `rescue block` there that logs the exception and increments the `error` count.

mejackreed:
Closes #99 This was causing an error on stage. Verified things work now.

### DIFF
--- a/lib/purl_finder.rb
+++ b/lib/purl_finder.rb
@@ -85,7 +85,11 @@ class PurlFinder
       druid = get_druid_from_file_path(line)
       if !druid.blank?
         IndexingLogger.info("indexing #{druid}")
-        result = Purl.save_from_public_xml(File.dirname(line)) # pass the directory of the file containing public
+        begin
+          result = Purl.save_from_public_xml(File.dirname(line)) # pass the directory of the file containing public
+        rescue => e
+          IndexingLogger.error("An error occurred while trying to save #{druid}.  \n#{e.message} \n#{e.backtrace.inspect}")
+        end
         result ? success += 1 : error += 1
         count += 1
       end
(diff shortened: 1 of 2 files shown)


lib/purl_finder.rb
```diff
@@ -85,7 +85,11 @@ def index_purls(params={})
       druid = get_druid_from_file_path(line)
       if !druid.blank?
         IndexingLogger.info("indexing #{druid}")
-        result = Purl.save_from_public_xml(File.dirname(line)) # pass the directory of the file containing public
+        begin
+          result = Purl.save_from_public_xml(File.dirname(line)) # pass the directory of the file containing public
+        rescue => e
+          IndexingLogger.error("An error occurred while trying to save #{druid}.  \n#{e.message} \n#{e.backtrace.inspect}")
+        end
```
drh-stanford: @mejackreed @lmcglohon just to be clear, does `result` become `false` if the exception is thrown? it's not clear to me that `error` gets incremented when an exception is raised from just reading the code. It looks like the spec does verify that `error` is getting set. thoughts?

mejackreed: Hm.. I see what you are saving. Maybe set to false above?

drh-stanford: Or maybe it's better to change the `Purl.save_from_public_xml` to thrown an exception (which may just be changing `.save` to `.save!`?) rather than return `true` or `false`? Then we could write something like this:

```ruby
begin
  Purl.save_from_public_xml(...)
  success += 1
rescue => e
  IndexingLogger.error(...)
  error += 1
end
```

mejackreed: Yeah that seems better 👍

PR #104, opened 2016-08-11
add PurlListener to watch for recent changes in published PURLs

### LINKED ISSUE #66: Add listener code to purl-fetcher
- [ ] implement listener process using gem (in #104)
- [ ] integrate listener process management with capistrano (in #104)
- [ ] develop entry point for listener to make index changes (in #104)
- [ ] document administrative use (where should this doc live?)
- [ ] implement changes to the `publish` robot (in #103)

Also see #61

### LINKED ISSUE #106: Fix Honeybadger exception logging on listener:restart
See https://app.honeybadger.io/projects/48916/faults/30715349

See #104

drh-stanford:
Paired with @eefahy on this PR.

This PR fixes #66 and fixes #106 by adding `PurlListener` to support incremental indexing. The listener is a daemon process and includes the `listener:start/stop/restart/status` tasks. A `cap deploy` and `cap deploy:restart` will restart the listener daemon.

See https://github.com/sul-dlss/purl-fetcher/issues/66#issuecomment-239324781 for a discussion of the basic algorithm.

**NOTE**: That this PR requires puppet changes before deployment. Right now only the `-dev` environment has those changes.

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -27,6 +27,7 @@ require 'capistrano/rvm'
 require 'whenever/capistrano'
 require 'capistrano/honeybadger'
 require 'capistrano/rails/migrations'
+require 'capistrano/passenger'
 
 # Loads custom tasks from `lib/capistrano/tasks' if you have any defined.
 Dir.glob('lib/capistrano/tasks/*.rake').each { |r| import r }
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -2,10 +2,6 @@
 
 set :output, '/dev/null'
 
-every 1.day do
- rake 'find:all' # does full scan of everything both changes and delete
-end
-
 every 15.minutes do
  rake 'find:deletes[15]' # scan .deletes directory for new deletes
 end
--- a/lib/purl_listener.rb
+++ b/lib/purl_listener.rb
@@ -1,0 +1,179 @@
+require 'benchmark'
+require 'fileutils'
+require 'listen'
+
+##
+# PurlListener supports incremental PURL updates via a "dropbox" folder
+# on the PURL filesystem. The implementation assumes that the PURL filesystem
+# is mounted via NFS and is a read-write mount.
+#
+# This is the basic algorithm:
+#
+# watches "dropbox" directory for filesystem events,
+# then on event:
+#     - logs that the listener process is active
+#     - rename `aa111bb2222` to `aa111bb2222.lock`
+#     - save to `Purl` model, and logs (to file and Honeybadger) any exceptions (then continues)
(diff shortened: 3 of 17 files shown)


Capfile
```diff
@@ -26,7 +26,8 @@ require 'dlss/capistrano'
 require 'capistrano/rvm'
 require 'whenever/capistrano'
 require 'capistrano/honeybadger'
-require 'capistrano/rails/migrations'
+require 'capistrano/rails'
```
mejackreed: Why the add back of assets precompile?

drh-stanford: I think Erin and I were experimenting with getting the `listener:restart` task run by cap deploy and this snuck in here. I'll restore it.

config/schedule.rb
```diff
@@ -2,10 +2,6 @@
 
 set :output, '/dev/null'
 
-every 1.day do
```
mejackreed: Why do we remove this now? I wonder could we make `find:all` idempotent and then still run that daily?

lib/purl_listener.rb
```diff
@@ -0,0 +1,179 @@
...
+  attr_reader :event_handler, :logger, :path, :pid_file
+
+  ##
+  # @param [Pathname] `path` -- the single "dropbox" directory
+  # @param [Pathname] `pid_file` -- the location of the pid file
+  # @param [Logger] `logger` -- place to log info/warn/errors
+  # @raise [ArgumentError] -- requires path to be a directory
+  def initialize(path = Pathname(PurlFetcher::Application.config.app_config['listener_path']),
+                 pid_file = Pathname(PurlFetcher::Application.config.app_config['listener_pid_file']),
+                 logger = IndexingLogger)
```
mejackreed: Is there a reason we use `IndexingLogger` here? Why not just the Rails logger?

mejackreed:
Looks good.. just a couple super minor questions/comments. Nothing blocking this though

config/schedule.rb
drh-stanford: Running it daily incurs the cost of the `find` command, but I don't think that's a big deal. The real question is whether the code handles unchanged `public` files correctly. So, I think it'd be another issue to fire off to at least verify that it does.

lib/purl_listener.rb
drh-stanford: Yes, `IndexingLogger` goes to the `log/indexing.log` file where as Rails logger goes to `log/production.log`. I think the idea is to keep all the "indexing" activity distinct from the API activity.

mejackreed: gotcha.. thx

PR #116, opened 2016-08-15
Always return sw preview

### LINKED ISSUE #111: docs/changes should return a true target for "preview"
probably should be configurable

mejackreed:
Closes #111


app/views/docs/changes.json.jbuilder
```diff
@@ -1,7 +1,7 @@
 json.changes @changes do |change|
   json.druid change.druid
   json.latest_change change.published_at.iso8601
-  json.true_targets change.release_tags.where(release_type: true).map(&:name) if change.release_tags.where(release_type: true).present?
+  json.true_targets change.release_tags.where(release_type: true).map(&:name) | Settings.ALWAYS_SEND_TRUE_TARGET
   json.false_targets change.release_tags.where(release_type: false).map(&:name) if change.release_tags.where(release_type: false).present?
```
drh-stanford: i don't think this works as intended if `Settings.ALWAYS_SEND_TRUE_TARGET` is `[]` or `nil`. can we enforce that there's at least one value in this constant array somehow? maybe in initializers/config.rb?

drh-stanford: I was thinking something like this in the initializer

```
fail unless Settings.ALWAYS_SEND_TRUE_TARGET.present?
```

mejackreed: Hm.. good point about that. I'll try to figure out what to do.

PR #132, opened 2016-08-17
remove :staging reference as its no longer valid

### LINKED ISSUE #18: Gemfile needs review
There's many gems in there that might be removed (i.e., jQuery?), and the groups also need cleanup (i.e., what is `:staging`?).

mejackreed:
Closes #18

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,7 @@ group :test do
   gem 'coveralls', require: false
 end
 
-group :staging, :production do
+group :production do
   gem 'mysql2'
 end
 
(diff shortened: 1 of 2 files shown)


lmcglohon:
@mejackreed Are there any gem require statements that can be removed?

mejackreed:
I'm not sure. I removed the frontend required gems in previous PR's. but didn't evaluate the entire app's gem requires.

lmcglohon:
@mejackreed Maybe this PR shouldn't close #18 until we have time to do a complete review?  Can it just reference #18 instead of close it?

mejackreed:
Hm. I think it closes the requirements from #18 unless I'm missing something. As far as I can tell, the Gemfile now includes only gems that are used in the app.

PR #136, opened 2016-08-19
ignore the /log/purl_finder file

mejackreed:


.gitignore
```diff
@@ -17,6 +17,7 @@
 /run/*.pid
 /db/*.sqlite3
 /db/*.sqlite3-journal
+/log/purl_finder
```
drh-stanford: this probably should be `/log/purl_finder*` so to ignore all the purl finder output files.

mejackreed: Thanks good catch.. added.

PR #137, opened 2016-08-19
adds a purl-api that is filterable by object type

mejackreed:


db/schema.rb
```diff
@@ -48,10 +48,13 @@
     t.datetime "created_at",   null: false
     t.datetime "updated_at",   null: false
     t.datetime "published_at"
+    t.string   "title"
```
drh-stanford: this needs to be `t.text` (or maybe a long string) as we have title data that's very large and it was generating insert exceptions.

mejackreed: Thanks good catch.. added.

Issue #157, opened 2016-08-24
[stage] Errno::ESTALE: Stale file handle


drh-stanford:
dor-services-app is getting these error messages:

```
Errno::ESTALE: Stale file handle @ utime_internal - /purl/recent_changes/pd507wn8117
```

It's possible we have a race condition between purl-fetcher's listener (renaming/deleting the file) and the publish robot (touching the file) on an NFS filesystem.

```
/usr/local/rvm/rubies/ruby-2.2.4/lib/ruby/2.2.0/fileutils.rb:1153:in `utime'
/usr/local/rvm/rubies/ruby-2.2.4/lib/ruby/2.2.0/fileutils.rb:1153:in `block in touch'
/usr/local/rvm/rubies/ruby-2.2.4/lib/ruby/2.2.0/fileutils.rb:1150:in `each'
/usr/local/rvm/rubies/ruby-2.2.4/lib/ruby/2.2.0/fileutils.rb:1150:in `touch'
/home/lyberadmin/dor-services-app/shared/bundle/ruby/2.2.0/gems/dor-services-5.10.0/lib/dor/models/publishable.rb:87:in `publish_not
ify_on_success'
/home/lyberadmin/dor-services-app/shared/bundle/ruby/2.2.0/gems/dor-services-5.10.0/lib/dor/models/publishable.rb:65:in `publish_met
adata'
/home/lyberadmin/dor-services-app/releases/20160817185754/lib/dor_services_app.rb:203:in `block (4 levels) in <class:DorServicesApi>
'
/home/lyberadmin/dor-services-app/shared/bundle/ruby/2.2.0/gems/grape-0.14.0/lib/grape/endpoint.rb:67:in `call'
```

It's possible that `FileUtils.touch` is not atomic over NFS -- see https://github.com/sul-dlss/dor-services/blob/master/lib/dor/models/publishable.rb#L87 -- and that we might need to switch to `File.write`

Also @eefahy we may need to disable `noac` on the purl-fetcher NFS mounts.

drh-stanford:
See http://stackoverflow.com/questions/8100574/creating-an-empty-file-in-ruby-touch-equivalent

drh-stanford:
Looks like we can tune the listener to delay the callbacks.

```
latency: 0.5                                    # Set the delay (**in seconds**) between checking for changes
                                                # default: 0.25 sec (1.0 sec for polling)

[...]

PR #158, opened 2016-08-24
slow down listener loop and delta between event and action

drh-stanford:
This PR is connected to #157. Configures the `listen` gem to slow down it's polling loop and the time delta between an event and callback action. This only reduces our exposure to the race condition described in #157.

### DIFF
--- a/lib/purl_listener.rb
+++ b/lib/purl_listener.rb
@@ -110,6 +110,8 @@ class PurlListener
       listener = Listen.to(path,
                            force_polling: true, # because on NFS mount
                            ignore: [/\.lock$/],
+                           latency: 60,         # wait between polling
+                           wait_for_delay: 5,   # see #157: slow down delta between change notice and action due to NFS
                            &event_handler)
       Process.daemon(true)
       @pid = $PROCESS_ID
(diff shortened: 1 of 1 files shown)


lib/purl_listener.rb
```diff
@@ -110,6 +110,8 @@ def listen_for_changes
       listener = Listen.to(path,
                            force_polling: true, # because on NFS mount
                            ignore: [/\.lock$/],
+                           latency: 60,         # wait between polling
+                           wait_for_delay: 5,   # see #157: slow down delta between change notice and action due to NFS
```
eefahy: I wonder if just configuring the `wait_for_delay` is enough to solve our problem?

drh-stanford: I've looked into `FileUtils.touch` quite a bit. It uses `File.utime` to change the mtime of the file *if* the file exists (which it doesn't), and so it uses `File.open(fn, 'a') {}` to create the file. Internally to ruby MRI that open/close calls utime_internal evidently. I don't think we can get more atomic than that for Ruby. Using `system("touch fn")` might be better as you're getting actual system calls at the lowest-level. Either way, I don't think we can do that much about the atomic-ness of the creating of the touch file -- I think `FileUtils.touch` is sufficient.

That said, the best thing we can do on the listener side is just slow it down -- `wait_for_delay` inserts a delay between the event (noticing that a file is created) and the callback action, as far as I can tell by digging into the `listen` gem.

As for the 5 second delay -- I dunno what the best value would be. I just chose it because it seems slow enough. 

The `latency` is a 60 second wait between polling -- that's the primary delay here. Again I chose it because it seems slow enough.

In production, we're getting at most a several touch files every minute. So I don't think the slowdown is too much for it. It does add clock time to watching an item go through the pipeline -- i.e., you have to wait a minute before it gets picked up by the listener. But since discovery dispatcher is run so infrequently, I don't think it affects the automated pipeline at all.

PR #160, opened 2016-08-25
rubocop updates

drh-stanford:
This PR puts us back into compliance with rubocop.


mejackreed:
Should we just add it to the CI build? Or add hound?

drh-stanford:
rubocop passes with this PR so I'm ok making it part of a successful build.

PR #180, opened 2016-10-11
True targets are not always so true

### LINKED ISSUE #175: If a purl is deleted, the corresponding record in the purl-fetcher database needs to be deleted
When a purl is deleted, as opposed to when the release data field is changed from true to false, the corresponding record in the purl-fetcher database needs to be removed.  For example, the purl for druid zc933cw0607 was deleted because the object was placed in the SDR graveyard. However, when the purl-fetcher API is called for the druid, the druid is returned with a true-target for SearchworksPreview. The purl-fetcher database should include only published purls so the database record should have been deleted. See attached files for the current corresponding argo record display and the result from the purl-fetcher API (purl-fetcher-prod.stanford.edu/purls/zc933cw0607)

zc933cw0607_argo

zc933cw0607_purl_fetcher

mejackreed:
Closes #175

### DIFF
--- a/app/models/purl.rb
+++ b/app/models/purl.rb
@@ -99,6 +99,8 @@ class Purl < ActiveRecord::Base
     purl = find_or_create_by(druid: druid) # either create a new druid record or get the existing one
     #  (in theory we should *always* have a previous druid here)
     purl.deleted_at = deleted_at.nil? ? Time.current : deleted_at
+    purl.release_tags.delete_all
+    purl.collections.delete_all
     purl.save
   end
 
--- a/app/models/purl.rb
+++ b/app/models/purl.rb
@@ -20,6 +20,22 @@ class Purl < ActiveRecord::Base
   }
 
   ##
+  # Return true targets with always values only if the object is not deleted in
+  # purl mount
+  # @return [Array]
+  def true_targets
+    return [] unless deleted_at.nil?
+    release_tags.where(release_type: true).map(&:name) | Settings.ALWAYS_SEND_TRUE_TARGET.to_a
+  end
+
+  ##
+  # Convenience method for accessing false targets
+  # @return [Array]
+  def false_targets
+    release_tags.where(release_type: false).map(&:name)
+  end
+
+  ##
   # Delete all of the collection assocations, and then add back ones from a
   # known valid list
   # @param [Array<String>] collections
--- a/spec/views/v1/collections/_collection.json.jbuilder_spec.rb
+++ b/spec/views/v1/collections/_collection.json.jbuilder_spec.rb
@@ -5,12 +5,11 @@ describe 'v1/collections/_collection.json.jbuilder' do
     render partial: 'v1/collections/collection', locals: { collection: Purl.first }
     expect(JSON.parse(rendered)).to include(
       'druid' => 'druid:ee111ff2222',
-      'true_targets' => ['SearchWorksPreview'],
       'catkey' => ''
     )
   end
-  it 'always returns "SearchWorksPreview"' do
-    render partial: 'v1/purls/purl', locals: { purl: Purl.first }
+  it 'always returns "SearchWorksPreview" for non deleted Purls' do
+    render partial: 'v1/collections/collection', locals: { collection: Purl.where(deleted_at: nil).first }
     expect(rendered).to match(/SearchWorksPreview/)
   end
 end
--- a/spec/views/v1/purls/_purl.json.jbuilder_spec.rb
+++ b/spec/views/v1/purls/_purl.json.jbuilder_spec.rb
@@ -7,13 +7,12 @@ describe 'v1/purls/_purl.json.jbuilder' do
       'collections' => ['druid:ff111gg2222'],
       'druid' => 'druid:bb111cc2222',
       'object_type' => 'item',
-      'true_targets' => ['SearchWorks', 'Revs', 'SearchWorksPreview'],
       'catkey' => '',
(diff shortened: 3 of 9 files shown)


app/models/purl.rb
```diff
@@ -20,6 +20,29 @@ class Purl < ActiveRecord::Base
   }
 
   ##
+  # Return true targets with always values only if the object is not deleted in
+  # purl mount
+  # @return [Array]
+  def computed_true_targets
```
drh-stanford: is there a case where `true_targets` should return the actual true targets of a deleted PURL? in other words, can you just put this logic into `true_targets`?

mejackreed: Yeah I'm not sure about that. I opted for the name `computed_true_targets` (not a great name I realize) so that someone wouldn't assume it just returned back whatever was there. What do you think?

drh-stanford: Maybe `true_targets_after_filter`? The name is challenging. I don't think I fully grasp the delete logic though.

peetucket: I like the idea of putting the logic in true_targets if that is the normal scenario of what is always going to be returned.  Even though it has the potential to confuse a consumer of the class (maybe they expected all), it reduces the risk of someone inadvertently returning the targets when they shouldn't (which seems like the normal case).  

Also, doesn't the update to the mark_deleted method method below means they should be cleaned up from now on?

app/models/purl.rb
```diff
@@ -20,6 +20,29 @@ class Purl < ActiveRecord::Base
...
+    return [] unless deleted_at.nil?
+    true_targets | Settings.ALWAYS_SEND_TRUE_TARGET.to_a
```
peetucket: Is that a bitwise or operator?  I've never used one and am curious - if there a reason to use that instead of the logical || operator?

mejackreed: Returns a unique set http://ruby-doc.org/stdlib-2.3.1/libdoc/set/rdoc/Set.html#method-i-7C

LynnMcRae:
I'm not sure I understand the question.  If true_targets are meant to express the true targets as expressed by the purl, and there is no purl, then there simply are no true targets. It's not a question that should be answered by filtering some deeper truth that DOR might know or the DB might (unnecessaily) remember regarding the history of the object ... no Purl means no true targets at all, or false ones for that matter, not even the supplied Preview one.  Purl Fetcher doesn't need this information; If you want to know more, ask Argo. Am I misunderstanding the question?

app/models/purl.rb
mejackreed: Thanks for the comments. I think I'm convinced to just add this into the `true_targets` method. Just simplifying the name, and taking into consideration @peetucket's great comment about how data going forward this shouldn't be a problem. Also conforms with @LynnMcRae 's comment.

mejackreed: Just pushed the revision.

app/models/purl.rb
peetucket: ah, basically a +

spec/views/v1/collections/_collection.json.jbuilder_spec.rb
```diff
@@ -5,12 +5,11 @@
...
     expect(JSON.parse(rendered)).to include(
       'druid' => 'druid:ee111ff2222',
-      'true_targets' => ['SearchWorksPreview'],
       'catkey' => ''
     )
   end
-  it 'always returns "SearchWorksPreview"' do
-    render partial: 'v1/purls/purl', locals: { purl: Purl.first }
+  it 'always returns "SearchWorksPreview" for non deleted Purls' do
+    render partial: 'v1/collections/collection', locals: { collection: Purl.find(6) }
```
drh-stanford: It's not clear to me why Purl `id=6` is the answer here. The test fixture for 6 says "6 should have no collection". Maybe add a comment?

spec/views/v1/purls/_purl.json.jbuilder_spec.rb
```diff
@@ -7,13 +7,12 @@
...
       'object_type' => 'item',
-      'true_targets' => ['SearchWorks', 'Revs', 'SearchWorksPreview'],
       'catkey' => '',
       'title' => 'Some test object'
     )
   end
-  it 'always returns "SearchWorksPreview"' do
-    render partial: 'v1/purls/purl', locals: { purl: Purl.first }
+  it 'always returns "SearchWorksPreview" for non deleted Purls' do
+    render partial: 'v1/purls/purl', locals: { purl: Purl.find(6) }
```
drh-stanford: Same as comment ^^^. Why is 6 the exemplar for a non-deleted collection?

spec/views/v1/collections/_collection.json.jbuilder_spec.rb
mejackreed: Ah.. ok, I can do that. The reason is that we really just want a `Purl` that isn't deleted. The other. Our "collections" are really just `Purl` with an association. And modifying the collection (to undelete) broke more tests.. so I opted for this easier path.

drh-stanford: Maybe do `Purl.where(deleted_at: nil).first`? Would something like that work? Part of my confusion is that the local variable is for `collection` not `purl` like below, but Purl 6 is an item not a collection object type.

mejackreed: Great suggestion. I pushed up a change to take your suggestion.

PR #186, opened 2016-10-17
return a 404 when in invalid collection druid is sent; add tests; rem…

### LINKED ISSUE #172: collections#show should return 404 for invalid druids
`[prod] /collections/index has undefined method 'druid' for nil:NilClass`

## Backtrace

    line 1 of [PROJECT_ROOT]/app/views/v1/purls/_purl.json.jbuilder: _app_views_v__purls__purl_json_jbuilder__1247116328238363603_29279540
    line 1 of [PROJECT_ROOT]/app/views/v1/collections/show.json.jbuilder: _app_views_v__collections_show_json_jbuilder___1489955878517212659_29357540

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/48916/faults/31113220)

peetucket:
…ove unnecessary _version.html view; add a simple default route that returns an "OK"

fixes #172


app/controllers/v1/collections_controller.rb
```diff
@@ -15,6 +15,7 @@ def index
     #
     def show
       @collection = Purl.find_by_druid(druid_param)
+      raise ActiveRecord::RecordNotFound if @collection.blank?
```
mejackreed: I think you can do this by just adding `!` to the end of `Purl.find_by_druid` 

See http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by-21

app/controllers/application_controller.rb
```diff
@@ -4,6 +4,10 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery unless: -> { request.format.json? }, with: :exception
 
+  def default
+    render text: 'ok'
```
mejackreed: Could we add a test for this action that maybe explains the need for it?

app/controllers/v1/collections_controller.rb
peetucket: cool, updated and squashed commit

app/controllers/application_controller.rb
peetucket: added a test and a comment to app controller --- not really needed per se but matches what some other API level apps we built do (and otherwise you get the generic Rails "welcome to your app" message if you happen to hit the root)

PR #188, opened 2016-10-25
remove references to indexing and replace with update and save

### LINKED ISSUE #161: Clean up "index" references

peetucket:
fixes #161 

renames a whole slew of methods, are they are consumers other than the rake task that need to be aware of this?

left a log file name as "indexing" since i wasn't sure if it was monitored

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 # purl-fetcher
 
-A web service app that queries PURL to return info needed for indexing or other purposes.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -48,7 +48,7 @@ bundle exec rake
 `GET /purls`
 
 #### Summary
-Purl Index route
+Purl Save route
```
mejackreed: Hmm.. I don't think this is right, can we switch back?

peetucket: yes, fixed

PR #193, opened 2016-10-27
Update to rails 5

cbeer:


drh-stanford:
This PR fails with `ActiveRecord::PendingMigrationError`.  The `ci` task does the `db:migrates`, so the change to the default target in the `Rakefile` in this PR breaks this.

peetucket:
master is now on rails5 from a separate PR

PR #197, opened 2017-03-17
[WIP] restart the listener on a regular schedule if it is down

peetucket:
Still up for discussion if this is useful or not...

### DIFF
--- a/lib/tasks/listener.rake
+++ b/lib/tasks/listener.rake
@@ -9,6 +9,17 @@ namespace :listener do
     end
   end
 
+  desc 'Restart the listener if it is down'
+  task :restart_if_down => :environment do |_t, args|
+    begin
+      listener = PurlListener.new
+      listener.start unless listener.running?
+    rescue SignalException => e
+      Honeybadger.notify(e)
+      puts "Listener status could not be checked or it could not be restarted: #{e.message}"
+    end
+  end
+
   desc 'Start the listener'
   task :start => :environment do |_t, args|
     begin
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -10,3 +10,8 @@ end
 every 1.hour do
   rake 'listener:recent_changes:process'
 end
+
+# ensures that the listener is still running (restart if not)
+every 8.hours do
+  rake 'listener:restart_if_down'
+end


lib/tasks/listener.rake
```diff
@@ -9,6 +9,17 @@ namespace :listener do
     end
   end
 
+  dsec 'Restart the listener if it is down'
```
drh-stanford: typo here broke the Travis build...

peetucket: sorry, fixed and rebased

drh-stanford:
There is a `restart` task already -- was that insufficient somehow? also, do we know why the listener went down? was the machine rebooted?

peetucket:
The current restart task stops the listener and starts it up again, not necessary if it is currently running.  The new one I wrote is a noop unless it is not currently running.

I couldn't find a cause for why it went down and am speculating that the listener had actually crashed.  If the process was indeed still running but was hung, this change would have no effect.

Separately, we also have a potential issue in that the discovery-dispatcher won't know to go back in time to find any purls that changed since the purl-fetcher hung.  One possible suggestion from Lynn is to have dispatcher keeping looking back in time until it hits a non-zero result just to be safe.

LynnMcRae:
"Separately, we also have a potential issue in that the discovery-dispatcher won't know to go back in time to find any purls that changed since the purl-fetcher hung."

My assumption is that the purl-fetcher never misbehaved...it simply continued to report recent findings from the index not having any way of knowing that a series of zero-results were due to lack of index updates (listener) rather than lack of purl updates.

peetucket:
I believe this assumption is correct.  Even if the listener is down or the process is hung, the purl-fetcher API will continue to respond with zero results to the dispatcher.

drh-stanford:
I guess my concern is that we're monitoring the listener process, see:

- https://purl-fetcher-prod/status/feature-listener-process
- https://github.com/sul-dlss/puppet/blob/production/hieradata/node/purl-fetcher-prod.stanford.edu.eyaml#L96-L100

did we actually get an alert? if not, something is either misconfigured or the code didn't work.

I'm not sure off-hand about whether discovery dispatcher needs to go back in time. The purl-fetcher listener is current with all the recent changes files so it's state should be good.

LynnMcRae:
"I'm not sure off-hand about whether discovery dispatcher needs to go back in time. The purl-fetcher listener is current with all the recent changes files so it's state should be good." 

It depends on whether purl-fetcher output is based on actual publication date, which can cause a gap, or on a timestamp that indicates when the listener updated the index, i.e. when the purl update can be detected by purl-fetcher. The latter would mean only listener needs to catch up.

LynnMcRae:
you can do a purl-fetcher search for times within the gap and get a result, e.g., 

https://purl-fetcher.stanford.edu/docs/changes?first_modified=2017-03-12T00:00:00Z&last_modified=2017-03-12T12:00:00Z

the "latest_change" date showing in the output corresponds to publication date. So it seems to be indexing, reporting, and dispatching based on publication date.

peetucket:
Indeed purl-fetcher stores the "published" date from public-xml in the database, which is later served back via the API.  It is not the timestamp of the purl file itself.

Where it gets the date and stores in the "published_at" database column:
https://github.com/sul-dlss/purl-fetcher/blob/master/app/models/purl.rb#L88

Where the API query runs against the date:
https://github.com/sul-dlss/purl-fetcher/blob/master/app/controllers/v1/docs_controller.rb#L8

and shown in json response:
https://github.com/sul-dlss/purl-fetcher/blob/master/app/views/v1/docs/changes.json.jbuilder#L3

peetucket:
But discovery dispatcher only looks back to the last time it ran, so we can still miss stuff?

see
https://github.com/sul-dlss/discovery-dispatcher/blob/master/lib/discovery_dispatcher/monitor.rb#L10 

which calls

https://github.com/sul-dlss/discovery-dispatcher/blob/master/lib/discovery_dispatcher/purl_fetcher_manager.rb#L9

which just goes to the database and gets the last run time

peetucket:
See #199  for a suggestion for using different dates when looking back in time

cbeer:
No listeners, no problems.

PR #199, opened 2017-03-21
update the purl fetcher changes api to use the updated_at column inst…

peetucket:
…ead of the published_at column when returning changes to avoid any gap issues for items published when purl-fetcher listener was down/hung

### DIFF
--- a/app/controllers/v1/docs_controller.rb
+++ b/app/controllers/v1/docs_controller.rb
@@ -5,7 +5,7 @@ module V1
     # API call to get a full list of all purls modified between two times
     def changes
       @changes = Purl.where(deleted_at: nil)
-                     .where(published_at: @first_modified..@last_modified)
+                     .where(updated_at: @first_modified..@last_modified)
                      .includes(:collections, :release_tags)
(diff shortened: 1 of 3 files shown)


mejackreed:
What constitutes a breaking change with this API?

peetucket:
As far as I know, the only consumer of the API is the dispatcher (for the indexing services), though you are correct in that if a consumer was really interested in the last_published date strictly, this change would change their results.  If we had a scenario like that, it would be easy to add a second API endpoint to provide that functionality.

mejackreed:
Gotcha. yeah I was just thinking about how we could codify all of the different api's, consumers and breaking changes.

peetucket:
This is a somewhat unusual API change in that it doesn't technically really break anything (call and response are identical), but it does affect what you get back (possibly returning more results than before if the listener had been down).  For indexing purposes, this is non-destructive (worst case, you index something again needlessly), but in other cases (e.g. a report), the additional results could be a problem.  

As you point out, perhaps the bigger message is that we don't have a standardized spot to document our APIs with a changelog of when things like this change.

PR #247, opened 2018-08-07
Allow doc changes and deletes to be filtered on their release tags

cbeer:
This should let us write indexing clients that go directly to purl fetcher to retrieve data, instead of waiting for discovery-dispatcher to push changes.


cbeer:
doc'ed. there are existing indexes on the release tags table that should cover this use.

PR #343, opened 2019-10-14
Update dependencies

sul-devops-team:


mejackreed:
Can we drop sprockets here all together?

cbeer:
I'm not sure what the happy path for non-sprockets asset delivery is; just in-line them in the template?

mejackreed:
Ah.. For some reason I thought this was an API only application. I guess its not anymore? I'm good w/ shipping then.

cbeer:
Nope; it has a spiffy home page with some stats on it.

PR #573, opened 2023-03-27
Install racecar (for working with kafka)

jcoyne:
This can be merged immediately, as it doesn't actually change any behavior.  Then, once we have enabled systemd to manage the background workers, and have configured Kafka, we can move work there.


thatbudakguy (approved):
Shortly after running `bundle exec racecar PurlUpdatesConsumer` I was seeing:
> Crashed: Rdkafka::Config::ConfigError: Configuration property "enable.ssl.certificate.verification" not supported in this build: OpenSSL not available at build time

Turns out it was [this issue](https://github.com/zendesk/racecar/issues/255), and the fix in [this comment](https://github.com/zendesk/racecar/issues/255#issuecomment-905766412) ultimately worked for me:
1. set `PKG_CONFIG_PATH` so that librdkafka knows where to find brew's openssl:
```sh
export PKG_CONFIG_PATH="/usr/local/opt/openssl@1.1/lib/pkgconfig"
```
2. reinstall librdkafka
```sh
brew reinstall librdkafka
```
3. rebuild the rdkafka gem's native extensions
```sh
bundle pristine rdkafka
```

Sharing in case this happens to anyone else who has openssl installed via brew. I imagine it won't affect linux targets like our VMs.

jcoyne:
@thatbudakguy I initially had the same problem but I fixed it with `brew install openssl && gem install rdkafka`

PR #586, opened 2023-03-30
Cache virtual object membership

### LINKED ISSUE #585: Index virtual object memberships
just as we index collections (but the inverse direction hasMembers)

jcoyne:
This is required for adding relatedItem entries on the public xml. Fixes #585

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -44,7 +44,7 @@ Metrics/AbcSize:
   Max: 30
 
 RSpec/MultipleExpectations:
-  Max: 7
+  Max: 9
 
 RSpec/MultipleMemoizedHelpers:
   Max: 6
--- a/app/models/cocina_data.rb
+++ b/app/models/cocina_data.rb
@@ -45,6 +45,15 @@ class CocinaData
     cocina_object.dro? ? cocina_object.structural.isMemberOf : []
   end
 
+  # Find virtual objects in argo via:
+  # https://argo-stage.stanford.edu/catalog?f%5Bhas_constituents_ssim%5D%5B%5D=has_constituents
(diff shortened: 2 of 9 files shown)


cbeer:
I think I don't understand this architecturally. Maintaining a parallel database of COCINA-derived data seems unfortunate. It still feels like SDR could send us everything we need to know up-front?

cbeer:
And, not just unfortunate, but doesn't this rely on us maintaining an accurate database? Async or parallel operations seem like we could get some inconsistent behavior.

It feels like it'd be nicer if infra maintained the only source of truth and hand over all the data we need in their original request.

jcoyne:
> It feels like it'd be nicer if infra maintained the only source of truth and hand over all the data we need in their original request.

They do hand over all the data.  We just need a way to look at this relationship inversely, just the same as we do with collection membership. This is just a cache of the data.

cbeer:
Maybe I don't understand what we're planning to use this data for. It seems to me, if it's data we need internally to construct the public xml, SDR should provide that to us to avoid any hint of race conditions, non-idempotent behavior, or need for synchronous processing.

cbeer:
> In this PR the access team takes on the responsibility for having an index of this data, just as we do for collection membership data

The difference with the collection membership data is that we're not trying to consume that synchronously.

Without infra handing us the data, won't we need to process a virtual object and the constituents in a well-defined order (make sure we've processed all the constituents before the virtual object?) or have some logic to update the virtual object whenever a constituent changes? That seems like at least as much of a pain.

I guess I don't know enough about how the Cocina is managed on the infra side; I guess I assumed there was just a big database they could ask?

PR #592, opened 2023-04-12
Remove old VERSION file and okcomputer version check

thatbudakguy:
Previously, we had been keeping the version in a file, which
got out of date (the app currently reports version 4.1.6).

<s>This changes to use the output of `git describe`, which combines
the latest version tag, number of commits since that tag, and
hash of most recent commit. If that fails, it just uses the full
commit hash.</s>

This removes the version check.


cbeer:
Do cap deploys have a `.git` directory?

thatbudakguy:
Apparently not 😢 they do have a REVISION file which `OkComputer::AppVersionCheck` will try to read from. Maybe that's good enough...? was hoping for a friendlier version output but at least it's not a lie

thatbudakguy:
Updated to just remove the version check.

jcoyne (approved):

PR #606, opened 2023-05-08
Do not process messages that have utf8m4 characters

jcoyne:


cbeer:
Instead of dropping them on the floor, could we just not store the title (or other metadata?) That way, downstream consumers could handle them until we get the root cause fixed.

jcoyne:
@cbeer thanks for the feedback. I've made that change.

thatbudakguy (approved):

PR #625, opened 2023-08-08
Extract a method for producing a reindex message

### LINKED ISSUE #622: Add the ability to send a kafka message from a Purl object
Before the kafka-ization of purl-fetcher, we could touch a `Purl` object and the data harvester on searchworks_traject_index would pick it up and put it in Kafka for indexing.

It would be helpful for administrative (and debugging) tasks if there was a method on the `Purl` class that enqueued a message on the `Settings.indexer_topic`. Otherwise, we need to rely on e.g. Andrew to republish data.

We anticipate wanting to use this feature to re-enqueue all Purls released to searchworks in preparation for the FOLIO cut-over to make sure we have a clean index.

jcoyne:
This will enable us to reindex all the purls.

```ruby
Purl.unscoped.find_in_batches.with_index do |group, batch|
  puts "Processing group ##{batch}"
  group.each(&:produce_indexer_log_message)
end
```

Or only for Searchworks:
```ruby
ReleaseTag.where(name: 'Searchworks').find_in_batches.with_index do |group, batch|
  puts "Processing group ##{batch}"
  group.each { |rt| rt.purl.produce_indexer_log_message }
end
```

Fixes #622

### DIFF
--- a/app/models/purl.rb
+++ b/app/models/purl.rb
@@ -47,6 +47,11 @@ class Purl < ApplicationRecord
     results
   end
 
+  # Sends a message to the indexer_topic, which will cause this object to be reindexed
+  def produce_indexer_log_message
+    Racecar.produce_sync(value: as_public_json.to_json, topic: Settings.indexer_topic, key: druid)
+  end
+
   def as_public_json
     data = if deleted?
              as_json(only: %i[druid])
--- a/README.md
+++ b/README.md
@@ -416,6 +416,23 @@ Name | Located In | Description | Required | Schema | Default
 
 ## Administration
 
+### Reindexing
+You can create Kafka messages that will cause all the Purls to be reindexed by doing:
+```ruby
+Purl.unscoped.find_in_batches.with_index do |group, batch|
+  puts "Processing group ##{batch}"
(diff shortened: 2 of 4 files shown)


app/models/purl.rb
```diff
@@ -47,6 +47,11 @@ def self.with_filter(filtering_params)
     results
   end
 
+  # Sends a message to the indexer_topic, which will cause this object to be reindexed
+  def produce_indexer_log_message
+    Racecar.produce_sync(value: as_public_json.to_json, topic: Settings.indexer_topic, key: druid)
```
cbeer, line 52: Back in the initial implementation, you tried to use `produce_sync` and removed it:

https://github.com/sul-dlss/purl-fetcher/pull/582/commits/664619592a3e2156d5ad69aa18016e5f02048cbd

The history is unclear, but do you remember what problem that fixed?

jcoyne: I was thinking that was a more idiomatic usage of Racecar.

thatbudakguy (approved):
This seems incredibly useful; can we get a code snippet (probably the one at the top of the PR) added to the README too?

PR #657, opened 2024-01-05
Upgrade to rails 7.1

jcoyne:


db/schema.rb
```diff
@@ -95,13 +94,26 @@
...
-    t.datetime "created_at", precision: 6, null: false
-    t.datetime "updated_at", precision: 6, null: false
+    t.datetime "created_at", precision: nil, null: false
+    t.datetime "updated_at", precision: nil, null: false
     t.index ["name", "purl_id"], name: "index_release_tags_on_name_and_purl_id", unique: true
     t.index ["purl_id"], name: "index_release_tags_on_purl_id"
     t.index ["release_type"], name: "index_release_tags_on_release_type"
   end
 
+  create_table "virtual_object_constituents", force: :cascade do |t|
```
corylown, line 104: @jcoyne where is this new table coming from? I'm not sure why there are schema changes in this pull request.

corylown (changes requested):

db/schema.rb
jcoyne: I'm not exactly sure.  I do know this happens when I run `db:migrate` even before upgrading rails.  See https://github.com/sul-dlss/purl-fetcher/pull/660

jcoyne: And if I don't include that change, the tests fail with:
```
SQLite3::SQLException: duplicate column name: data_type
```

jcoyne: It must have been a bad db state.  It goes away if I do `bin/rails db:drop db:create db:migrate`

corylown (approved):

PR #683, opened 2024-04-15
Simplify status scope and use it consistently

jcoyne:


app/models/purl.rb
```diff
@@ -21,7 +21,7 @@ class Purl < ApplicationRecord
   }
 
   scope :status, lambda { |status|
-    case status['status']
+    case status
     when 'deleted'
       where.not deleted_at: nil
     when 'public'
```
thatbudakguy, line 27: imo it's kind of weird that we call this "public", but 🤷🏻

thatbudakguy (approved):

PR #688, opened 2024-04-26
Decouple release tags from Cocina

jcoyne:
This will allow us to eventually remove release tags from the Cocina model. This supports versioning changes

### DIFF
--- a/app/controllers/v1/purls_controller.rb
+++ b/app/controllers/v1/purls_controller.rb
@@ -18,7 +18,15 @@ module V1
                 retry
               end
 
-      Racecar.produce_sync(value: cocina_object.to_json, key: druid_param, topic: "purl-updates")
+      # Get the actions from the release tags on the cocina model. In the near future,
+      # the releaseTags will be removed from Cocina.
+      actions = { index: [], delete: [] }.tap do |releases|
+        cocina_object.administrative.releaseTags.each do |tag|
+            releases[tag.release ? :index : :delete] << tag.to
+        end
+      end
+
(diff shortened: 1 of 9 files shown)


app/controllers/v1/purls_controller.rb
```diff
@@ -18,7 +18,15 @@ def update
                 retry
               end
 
-      Racecar.produce_sync(value: cocina_object.to_json, key: druid_param, topic: "purl-updates")
+      # Get the actions from the release tags on the cocina model. In the near future,
+      # the releaseTags will be removed from Cocina.
+      actions = { index: [], delete: [] }.tap do |releases|
+        cocina_object.administrative.releaseTags.each do |tag|
+            releases[tag.release ? :index : :delete] << tag.to
```
mjgiarlo, line 25: Indentation looks excessive here but I guess if rubocop doesn't care, I'm not too fussed

app/models/purl.rb
```diff
@@ -97,10 +97,10 @@ def false_targets
   end
 
   # add the release tags, and reuse tags if already associated with this PURL
-  def refresh_release_tags(releases)
-    [true, false].each do |type|
-      releases[type.to_s.to_sym].sort.uniq.each do |release|
-        release_tags << ReleaseTag.for(self, release, type)
+  def refresh_release_tags(actions)
+    ['index', 'delete'].each do |type|
```
mjgiarlo, line 101: And we want string keys here since the hash is being converted from a JSON string, right?

mjgiarlo (approved):

app/models/purl.rb
jcoyne: exactly.

PR #689, opened 2024-04-26
Create a route for transfering release tags

jcoyne:


justinlittman (commented):
How does DSA know what to delete?

jcoyne:
@justinlittman presumably user input?

justinlittman:
DSA knows where an item should be released; PURL Fetcher should know where it actually has been released. Assuming that DSA knows where an item actually has been released to seems like an opportunity for systems to get out of sync.

In other words, I'd suggest that DSA should tell PURL Fetcher the expected release state; PURL Fetcher should determine what actions are necessary to achieve that state.

jcoyne:
@justinlittman Yes, isn't that exactly what we have here?

justinlittman:
Doesn't this API change require DSA to provide the index and delete actions (instead of just a list of release targets)?

jcoyne:
@justinlittman "index" is just an improved name (IMHO) for "true_targets" and "delete" is the improved name for "false_targets"

PR #707, opened 2024-05-02
Adds jwt authentication.

### LINKED ISSUE #706: Add shared token authentication for API requests
The requests that modify data should be behind some sort of authentication.

Similar to DSA https://github.com/sul-dlss/dor-services-app?tab=readme-ov-file#authentication

justinlittman:
refs #706

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -67,3 +67,5 @@ gem 'cocina-models'
 gem "racecar"
 
 gem 'whenever', require: false
+
+gem 'jwt' # json web token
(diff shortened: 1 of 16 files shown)


jcoyne (approved):

jcoyne:
@justinlittman Do we have tokens generate on stage and prod?  Do we need to update dor-services-app to use this token when publishing?

justinlittman:
No, we don't have tokens generated yet. A new version of the client will need to be cut and tokens / changes made to DSA.

I've left the underlying ticket open to continue that work.

PR #710, opened 2024-05-02
Purge active storage attachments weekly.

### LINKED ISSUE #705: write a cron job that deletes activestorage files after 1 week
This will allow us to test uploading files, but not overwhelm disk storage.

justinlittman:
closes #705

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -1,0 +1,25 @@
+# Use this file to easily define all of your cron jobs.
+#
+# It's helpful, but not entirely necessary to understand cron before proceeding.
+# http://en.wikipedia.org/wiki/Cron
+
+# Example:
+#
+# set :output, "/path/to/my/cron_log.log"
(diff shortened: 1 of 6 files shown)


config/schedule.rb
```diff
@@ -0,0 +1,25 @@
...
+#
+# every 4.days do
+#   runner "AnotherModel.prune_old_records"
+# end
+
+# Learn more: http://github.com/javan/whenever
+
+# Purging ActiveStorage attachments during development to avoid filling storage.
+every 7.days do
+  runner "ActiveStorage::Attachment.all.each {|attachment| attachment.purge}"
```
jcoyne, line 24: I don't think we make any attachments.  It's the Blobs we want to purge.

jcoyne (changes requested):

jcoyne (approved):

PR #724, opened 2024-05-08
Prototype Public Xml generation from JSON

jcoyne:


app/models/purl.rb
```diff
@@ -1,8 +1,10 @@
-class Purl < ApplicationRecord # rubocop:disable Metrics/ClassLength
+class Purl < ApplicationRecord
   has_and_belongs_to_many :collections
   has_many :constituent_memberships, inverse_of: :parent, dependent: :destroy
   has_many :constituents, -> { order 'constituent_memberships.sort_order' },
            through: :constituent_memberships, source: :child
+  has_many :parent_memberships, class_name: 'ConstituentMembership', inverse_of: :child, dependent: :destroy
```
justinlittman, line 6: Is this unrelated?

app/models/purl.rb
```diff
@@ -90,7 +92,7 @@ def deleted?
   ##
   # Release tags where the value is true or is one of the default targets. If the object has been deleted, it retuns blank.
   # This is consumed by
-  # https://github.com/sul-dlss/searchworks_traject_indexer/blob/64359399e8f670ed414b1c56c648dc9b95ad6bad/lib/traject/readers/kafka_purl_fetcher_reader.rb#L26 # rubocop:disable Layout:LineLength
+  # https://github.com/sul-dlss/searchworks_traject_indexer/blob/64359399e8f670ed414b1c56c648dc9b95ad6bad/lib/traject/readers/kafka_purl_fetcher_reader.rb#L26:LineLength
```
justinlittman, line 95: Something seems to have gone wrong here.

app/services/cocina_object_store.rb
```diff
@@ -0,0 +1,13 @@
+class CocinaObjectStore
+  def self.find(druid)
+    path = "#{purl_druid_path(druid)}/cocina.json"
+    data_hash = JSON.parse(File.read(path))
+    Cocina::Models.build(data_hash)
```
justinlittman, line 5: So let's suppose we're rolling out a breaking cocina change and re-publishing all items. What happens if we publish an object which requires loading objects that have not yet been republished? That is, are we requiring a republishing order?

app/models/purl.rb
jcoyne: No, this is how we get the constituents for virtual objects

justinlittman: Gotcha.

app/services/cocina_object_store.rb
jcoyne: Only in the case of Virtual objects, where the constituents need to be published before the object that is composed of the constituents.

justinlittman (approved):

PR #743, opened 2024-05-17
Add MODS endpoint for Argo

### LINKED ISSUE #742: Add API for displaying mods given some Cocina

jcoyne:
Allows Argo to dump XML creation.  It can call this as a service.

Fixes #742

### DIFF
--- a/app/controllers/v1/mods_controller.rb
+++ b/app/controllers/v1/mods_controller.rb
@@ -1,0 +1,11 @@
+module V1
+  class ModsController < ApplicationController
+    # Given a POST body with cocina, transform it to MODS xml. Used by Argo to preview metadata
+    def create
+      public_cocina = Cocina::Models.build(params[:mods].to_unsafe_h)
+      desc_metadata_service = Publish::PublicDescMetadataService.new(public_cocina, [])
+      desc_md_xml = desc_metadata_service.ng_xml(include_access_conditions: false)
+      render xml: desc_md_xml.to_xml
+    end
+  end
+end
--- a/config/initializers/inflections.rb
+++ b/config/initializers/inflections.rb
@@ -3,12 +3,9 @@
 # Add new inflection rules using the following format. Inflections
 # are locale specific, and you may define rules for as many different
 # locales as you wish. All of these examples are active by default:
-# ActiveSupport::Inflector.inflections(:en) do |inflect|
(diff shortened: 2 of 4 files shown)


app/controllers/v1/mods_controller.rb
```diff
@@ -0,0 +1,11 @@
+module V1
+  class ModsController < ApplicationController
+    # Given a POST body with cocina, transform it to MODS xml. Used by Argo to preview metadata
+    def create
+      public_cocina = Cocina::Models.build(params[:mod].to_unsafe_h)
```
mjgiarlo, line 5: Is `:mod` a typo?

spec/requests/v1/mods_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.describe "Mods export" do
```
mjgiarlo, line 3: How is this test passing if the controller is using a `:mod` parameter which appears to be missing here?

app/controllers/v1/mods_controller.rb
jcoyne: Nope, it's how rails is automatically de-pluralizing "mods"

spec/requests/v1/mods_spec.rb
jcoyne: Rails magick ✨

jcoyne: Rails assigns the body to a parameter derived from the controller name.

mjgiarlo: TIL

config/initializers/inflections.rb
```diff
@@ -3,12 +3,9 @@
...
 # locales as you wish. All of these examples are active by default:
-# ActiveSupport::Inflector.inflections(:en) do |inflect|
-#   inflect.plural /^(ox)$/i, "\\1en"
-#   inflect.singular /^(ox)en/i, "\\1"
-#   inflect.irregular "person", "people"
-#   inflect.uncountable %w( fish sheep )
-# end
+ActiveSupport::Inflector.inflections(:en) do |inflect|
+  inflect.uncountable %w( mods )
+end
```
mjgiarlo, line 8: Old_Mods_photo

mjgiarlo (approved):

PR #745, opened 2024-05-20
Prototype publish api

### LINKED ISSUE #718: Prototype new publish API
It should:
- [ ] shelve files uploaded to ActiveStorage
- [ ] put the metadata into `/purl`
- [ ] handle the delete of a file that was previously published
- [ ] handle a file that is already present from a previous publish.

This would be blocked by #719, but hopefully we won't need to handle the transfer of public xml, we can just create it from the cocina.  So, build this assuming public XML is a solved problem and that we only need to get cocina.json in place.

aaron-collier:
Closes #718 

This is a basic, **prototype**, publish API. 

This:
- [x] moves files that have a valid ActiveStorage signed_id to the stacks path for the druid
- [x] Remoes files from the stacks path for the druid if the filename is not in cocina
- [x] Writes the cocina.json file to the purl path
- [x] If an updated file is uploaded to active storage the existing file is overwritten

This is mostly pulled from SDR api so far.

### DIFF
--- a/app/controllers/v1/resources_controller.rb
+++ b/app/controllers/v1/resources_controller.rb
@@ -40,9 +40,7 @@ module V1
     # Write the cocina object to the Purl druid path as cocina.json
     # return [String] the path to the written cocina.json file
     def write_purl
-      cocina_json_path = File.join(purl_druid_path, 'cocina.json')
       FileUtils.mkdir_p(purl_druid_path) unless File.directory?(purl_druid_path)
-      File.write(cocina_json_path, cocina_object.to_json)
 
       purl = begin
         Purl.find_or_create_by(druid:)
@@ -50,9 +48,24 @@ module V1
         retry
       end
 
+      write_public_cocina
+      write_public_xml
+
       Racecar.produce_sync(value: { cocina: cocina_object, actions: nil }.to_json, key: druid, topic: "purl-updates")
 
       purl
+    end
+
+    def write_public_cocina
+      File.write(File.join(purl_druid_path, 'cocina.json'), cocina_object.to_json)
+    end
+
+    def write_public_xml
+      File.write(File.join(purl_druid_path, 'public.xml'), public_xml)
+    end
+
+    def public_xml
+      Publish::PublicXmlService.new(public_cocina: cocina_object, thumbnail_service: ThumbnailService.new(cocina_object)).to_xml
     end
 
     CREATE_PARAMS_EXCLUDE_FROM_COCINA = %i[action controller resource].freeze
--- a/app/controllers/v1/resources_controller.rb
+++ b/app/controllers/v1/resources_controller.rb
@@ -1,0 +1,138 @@
+# frozen_string_literal: true
+
+module V1
+  class ResourcesController < ApplicationController
+    include Authenticated
+
+    class BlobError < StandardError; end
+
+    before_action :check_auth_token, only: %i[create]
+
+    # POST /resource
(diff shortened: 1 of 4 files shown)


app/controllers/v1/resources_controller.rb
```diff
@@ -0,0 +1,304 @@
...
+
+require 'base64'
+
+module V1
+  # rubocop:disable Metrics/ClassLength
+  class ResourcesController < ApplicationController
+    include Authenticated
+
+    class BlobError < StandardError; end
+    class GlobusNotFoundError < StandardError; end
```
jcoyne, line 11: We won't need Globus

app/controllers/v1/resources_controller.rb
```diff
@@ -0,0 +1,304 @@
...
+
+    before_action :check_auth_token, only: %i[create update]
+    before_action :validate_version, only: %i[create update]
+
+    GLOBUS_PREFIX = 'globus://'
+
+    # GET /resource/:id
+    def show
```
jcoyne, line 19: We don't need the show endpoint.

app/services/cocina_version_validator.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# Checks that the major and minor cocina versions match.
+class CocinaVersionValidator
```
jcoyne, line 4: I don't think we want this.  It was important in SDR-API where we had less control over the clients, but here we only have one party (DSA) sending requests.

app/controllers/v1/resources_controller.rb
```diff
@@ -0,0 +1,304 @@
...
+    end
+
+    # POST /resource
+    # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/AbcSize
+    def create
+      begin
+        @druid = cocina_object.externalIdentifier
+        shelve_files
+        location = write_purl
```
jcoyne, line 36: You must also stick this stuff in here:
https://github.com/sul-dlss/purl-fetcher/blob/main/app/controllers/v1/purls_controller.rb#L16-L22

app/controllers/v1/resources_controller.rb
```diff
@@ -0,0 +1,141 @@
+# frozen_string_literal: true
+
+require 'base64'
+require 'digest'
```
jcoyne, line 4: Where is this used?

app/controllers/v1/resources_controller.rb
```diff
@@ -0,0 +1,141 @@
...
+    def stacks_druid_path
+      DruidTools::PurlDruid.new(druid, Settings.filesystems.stacks_root).path
+    end
+
+    # Write the cocina object to the Purl druid path as cocina.json
+    # return [String] the path to the written cocina.json file
+    def write_purl
+      cocina_json_path = File.join(purl_druid_path, 'cocina.json')
+      FileUtils.mkdir_p(purl_druid_path) unless File.directory?(purl_druid_path)
+      File.write(cocina_json_path, cocina_object.to_json)
```
jcoyne, line 48: Just noting that we need to do the public XML here too.

app/controllers/v1/resources_controller.rb
aaron-collier: @jcoyne good catch, cruft. This was from early debugging. Removed. Thanks for the heads up.

jcoyne (approved):

PR #750, opened 2024-05-21
Write the purl-fetcher meta information to the purl file system

cbeer:


app/controllers/v1/resources_controller.rb
```diff
@@ -49,6 +49,7 @@ def write_purl
       end
 
       write_public_cocina
+      write_public_purl_metadata(purl)
```
jcoyne, line 52: This should go into the `ReleasedController#update` method.

app/controllers/v1/resources_controller.rb
```diff
@@ -60,6 +61,10 @@ def write_public_cocina
       File.write(File.join(purl_druid_path, 'cocina.json'), cocina_object.to_json)
     end
 
+    def write_public_purl_metadata(purl)
+      File.write(File.join(purl_druid_path, 'meta.json'), purl.as_public_json.to_json)
```
jcoyne, line 65: I think we should reconsider the format of this file if we decide to go down this path.  I find that `true_targets` and `false_targets` to not be super helpful without context. So how about something like `index` and `delete`.  That's what we're using in the purl-fetcher API to create release tags: https://github.com/sul-dlss/purl_fetcher-client/blob/main/lib/purl_fetcher/client/release_tags.rb#L34

jcoyne (changes requested):

app/controllers/v1/resources_controller.rb
cbeer: I think we'll want it in both places, to ensure we always have meta information even for newly released/never released objects, right?

app/controllers/v1/resources_controller.rb
cbeer: I don't have an opinion about the format, although I do wonder if we should decouple changing the format from this change.

app/controllers/v1/resources_controller.rb
```diff
@@ -12,7 +12,7 @@ class ResourcesController < ApplicationController
     def create
       begin
         UpdateStacksFilesService.new(@cocina_object).write!
-        UpdatePurlMetadataService.new(@cocina_object).write!
+        UpdatePurlMetadataService.new(@cocina_object, @purl).write!
```
jcoyne, line 15: I think this should be `only: [:cocina, :xml]`. We don't want this action to touch `meta.json`

jcoyne: DOR services app automatically sends the Released#create API call right after the Purl#updates call, so this should be taken care of by that: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/publish/metadata_transfer_service.rb#L77-L85

cbeer: And we don't want to change dor-services to send that information with the cocina data?

Also, is there any harm in writing the same file twice? I guess to me, it seems like there's some benefits to being proactive on the purl-fetcher side in order to avoid half-baked objects.

jcoyne: >

> And we don't want to change dor-services to send that information with the cocina data?

Versioned vs unversioned data.  We are splitting purl-fetcher into two APIs, one that says "put these files and data in place" (previously publish/shelve) and one that says "send this stuff on to indexers"

> Also, is there any harm in writing the same file twice? I guess to me, it seems like there's some benefits to being proactive on the purl-fetcher side in order to avoid half-baked objects.

Yes, I think it's confusing if one of the endpoints does publish/shelve/release and one just does release. We're working to simplify the process into two distinct services.

cbeer: 🤷‍♂️ ok, as long as it makes sense to you.

cbeer, line 15:
```suggestion
        UpdatePurlMetadataService.new(@cocina_object, @purl).write! only: [:cocina, :xml]
```

app/controllers/v1/released_controller.rb
```diff
@@ -20,8 +20,8 @@ def update
 
       # add the release tags, and reuse tags if already associated with this PURL
       purl.refresh_release_tags(actions)
+      UpdatePurlMetadataService.new(nil, purl).write!(only: [:meta])
```
jcoyne, line 23: Can we move the meta writing into its own class? I think that makes it simpler than having a bunch of conditionals around the `only:` parameter.  Please restore the indexer message here, or create a `ReleaseService` which combines writing the meta.json + kafka message.

cbeer: It seems like you have a very clear notion of how you'd like this work done, and I'm not sure I share that clarity. I'll just close this PR and you can pick up the underyling ticket.

PR #752, opened 2024-05-22
Extract services from ResourcesController to write to stacks + purl filesystems

cbeer:


jcoyne (approved):
Thank you! This is much easier to wrap one's head around.

PR #754, opened 2024-05-23
Specify content directory when migrating files into OCFL structure

mjgiarlo:


app/services/ocfl_migrator.rb
```diff
@@ -20,7 +20,7 @@ def migrate
   attr_reader :druid
 
   def builder
-    @builder ||= OCFL::Object::DirectoryBuilder.new(object_root: ocfl_druid_path, id: druid)
+    @builder ||= OCFL::Object::DirectoryBuilder.new(object_root: ocfl_druid_path, id: druid, content_directory: 'data')
```
jcoyne, line 23: Why are we using `data` instead of the default (`content` I believe)?

mjgiarlo: I had a vague recollection that we had discussed this, but now that I search slack, I see that you thought we didn't need to override this (for alignment with moab practices). 😵 I will undo this change. It's fine to use the default.

PR #760, opened 2024-05-23
Update the local cocina data in resources#create

cbeer:
... and use the `Purl` model instead of cocina in the various services.


jcoyne (changes requested):
Doesn't app/consumers/purl_updates_consumer.rb need to be updated too? It references PurlCocinaUpdater

app/controllers/v1/resources_controller.rb
```diff
@@ -10,9 +10,11 @@ class ResourcesController < ApplicationController
 
     # POST /resource
     def create
+      PurlCocinaUpdater.new(@purl, @cocina_object).update
```
jcoyne, line 13: If we're doing this in process now, should this be removed? https://github.com/sul-dlss/purl-fetcher/blob/27ad4168a007c3c7162178ea29858850c440d4d1/app/services/update_purl_metadata_service.rb#L20

jcoyne (changes requested):

app/controllers/v1/resources_controller.rb
cbeer: My understanding is we're not yet using the resources controller, so removing that would break the existing processing. I don't think it's particularly harmful to process the data twice and the kafka topic with cocina data may be useful for indexing in the future (mirroring the way we're processing FOLIO's data).

jcoyne (approved):

PR #762, opened 2024-05-24
Write cocina from PurlsController#update.

### LINKED ISSUE #741: Publish the generated cocina
Do this for both the new `ResourcesController#create`  and for the legacy `PurlsController#update` API endpoints. 
Generate the cocina and move it to the /purl mount.  Once this is complete, DSA can stop sending it.

When this is done, create a DSA ticket to remove that functionality from DSA.

justinlittman:
closes #741


app/services/public_xml_writer.rb
```diff
@@ -1,5 +1,9 @@
 class PublicXmlWriter
-  def self.write(public_cocina, output_path)
+  def self.write(public_cocina, output_path: nil)
+    output_path ||= File.join(DruidTools::PurlDruid.new(public_cocina.externalIdentifier, Settings.filesystems.purl_root).path, 'public.xml')
```
jcoyne, line 3: Can we just pass this in and avoid any conditional?

app/services/public_xml_writer.rb
```diff
@@ -1,5 +1,9 @@
...
+    output_dir = File.dirname(output_path)
+    FileUtils.mkdir_p(output_dir) unless File.directory?(output_dir)
```
jcoyne, line 5: In which case wouldn't this exist?  We shouldn't need to do this.

jcoyne (changes requested):

app/services/update_purl_metadata_service.rb
```diff
@@ -1,41 +1,8 @@
...
-  end
-
-  def public_xml
-    Publish::PublicXmlService.new(public_cocina: cocina_object, thumbnail_service: ThumbnailService.new(cocina_object)).to_xml
-  end
-
-  def druid
-    cocina_object.externalIdentifier
+  def self.write!(cocina_object)
+    PublicCocinaWriter.write(cocina_object)
```
jcoyne, line 4: I'd prefer we have the directory building happening here rather than duplicating it in each of the other writers.

app/services/update_purl_metadata_service.rb
```diff
@@ -1,41 +1,8 @@
...
   # Write the cocina object to the Purl druid path as cocina.json
-  def write!
-    FileUtils.mkdir_p(purl_druid_path) unless File.directory?(purl_druid_path)
-
-    write_public_cocina
-    write_public_xml
-    write_kafka_message
-  end
-
-  def write_kafka_message
```
jcoyne, line 17: I think maintaining this as a method helps to convey what this is doing.

PR #764, opened 2024-05-24
[refactor] extract PublicXmlWriter and GeneratedXmlTester

jcoyne:


app/services/public_xml_writer.rb
```diff
@@ -0,0 +1,7 @@
+class PublicXmlWriter
+  def self.write(_cocina, output_path)
```
justinlittman, line 2: Looks like the method args aren't correct.

justinlittman (changes requested):

justinlittman (approved):

PR #772, opened 2024-05-24
Write cocina from PurlsController#update.

### LINKED ISSUE #741: Publish the generated cocina

justinlittman:
closes #741


app/consumers/purl_updates_consumer.rb
```diff
@@ -11,6 +11,8 @@ def process(message)
     purl = Purl.find_by!(druid: cocina_object.externalIdentifier)
     PurlCocinaUpdater.new(purl, cocina_object).update
 
+    write_public_cocina(cocina_object, purl.purl_druid_path)
+
     test_public_xml_generation(cocina_object)
```
jcoyne, line 16: I don't think there's any point in continuing to test. I feel I may have been remiss in failing to point this out in the issue.  This is now "Swap testing for actually writing the file"

jcoyne (changes requested):

jcoyne (approved):

PR #774, opened 2024-05-24
Move stacks directory check outside of the loop

jcoyne:


mjgiarlo (approved):

app/services/update_stacks_files_service.rb
```diff
@@ -38,21 +38,18 @@ def check_signed_ids
...
     end
   end
 
-  def make_shelving_dir(shelving_path)
-    shelving_dir = File.dirname(shelving_path)
-    return if File.directory?(shelving_dir)
-
-    FileUtils.mkdir_p(shelving_dir)
+  def make_shelving_dir(shelving_dir)
+    File.directory?(shelving_dir) || FileUtils.mkdir_p(shelving_dir)
```
mjgiarlo, line 52: Won't `FileUtils.mkdir_p(shelving_dir)` internally no-op if `File.directory?(shelving_dir)` returns `true`? In other words, can we ditch the check? Surprised this didn't trip Rubocop's `Lint/NonAtomicFileOperation` cop.

jcoyne: I think so. I'll try that.

PR #790, opened 2024-05-31
Write to ocfl.

### LINKED ISSUE #721: Write OCFL objects
Always just overwrite v1 for now.  Keep placing files in the legacy /purl and /stacks as well.

blocked by #718

justinlittman:
closes #721

### DIFF
--- a/app/services/update_ocfl_service.rb
+++ b/app/services/update_ocfl_service.rb
@@ -1,0 +1,103 @@
+class UpdateOcflService
+  class BlobError < StandardError; end
+
+  COCINA_JSON = 'cocina.json'.freeze
+  PUBLIC_XML = 'public.xml'.freeze
+
+  def self.write!(...)
+    new(...).write!
+  end
+
+  def initialize(purl, file_uploads_map)
+    @purl = purl
+    @file_uploads_map = file_uploads_map
+    @cocina_tempfile = Tempfile.new(COCINA_JSON)
+    @public_xml_tempfile = Tempfile.new(PUBLIC_XML)
+  end
+
+  def write!
+    add_files
+    add_cocina_json
+    add_public_xml
+    remove_files
+    ocfl_version.save
+  ensure
+    cocina_tempfile.close
+    cocina_tempfile.unlink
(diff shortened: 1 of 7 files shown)


app/services/update_ocfl_service.rb
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  # Copy the files from ActiveStorage to ocfl
+  def add_files
+    file_uploads_map.each do |filename, signed_id|
+      blob = blob_for_signed_id(signed_id, filename)
+      blob_path = ActiveStorage::Blob.service.path_for(blob.key)
+
+      dir_name = File.dirname(filename)
+      builder.copy_file(blob_path, destination_path: dir_name == '.' ? nil : dir_name)
```
jcoyne, line 47: Should we move this conditional upstream?

app/services/update_ocfl_service.rb
```diff
@@ -0,0 +1,94 @@
...
+  def blob_for_signed_id(signed_id, filename)
+    file_id = ActiveStorage.verifier.verified(signed_id, purpose: :blob_id)
+    ActiveStorage::Blob.find(file_id)
+  rescue ActiveRecord::RecordNotFound
+    raise BlobError, "Unable to find upload for #{filename} (#{signed_id})"
+  end
+
+  # Remove files from the ocfl that are not in the cocina object
+  def remove_files
+    ocfl_files_to_delete = directory.versions[directory.head].file_names.reject { |ocfl_file| cocina_filenames.include?(ocfl_file) }
```
jcoyne, line 61: could we look at `builder.version.state.values.flatten` instead of `directory.versions[directory.head].file_names` ?


I don't like that we have both a handle on `builder` and `directory`

app/services/update_ocfl_service.rb
```diff
@@ -0,0 +1,94 @@
...
+    ocfl_files_to_delete.each do |ocfl_file|
+      builder.delete_file(ocfl_file)
```
jcoyne, line 63: Delete file takes a SHA512, not a name.

PR #809, opened 2024-06-24
Delete files when marking a purl dark

### LINKED ISSUE #806: New publish/shelve API does not delete files from Stacks when files are changed to "shelve=no"
**Describe the bug**
When an item's rights are changed to dark, the publish and shelve settings are set to "no". That means that any files on Stacks should be removed when the accessionWF "shelve" step runs after the shelve settings were changed to no.

When I make an item dark in stage, I see the files are all left behind in Stacks.

**User Impact**
Files don't get deleted when they should and are left without anyone knowing they're there. This is one of the ways the state of the Stacks filesystem falls out of sync with Cocina.

**To Reproduce**
1. Accession an item that has files and object rights set to something other than dark.
2. After accessioning, open a new version and set its rights to dark.
3. Close the item.
4. Go to the Stacks filesystem and see that the files are still there.

**Expected behavior**
Files should be deleted when an item is made dark. More specifically, if an item has files with "shelve=no", those files on Stacks should get deleted if they were previously on Stacks.

jcoyne:
Fixes #806

### DIFF
--- a/app/controllers/v1/purls_controller.rb
+++ b/app/controllers/v1/purls_controller.rb
@@ -3,6 +3,7 @@ module V1
     include Authenticated
 
     before_action :check_auth_token, only: %i[update destroy]
+    before_action :find_purl, only: %i[update destroy]
 
     # Show the public json for the object. Used by purl to know if this object should be indexed by crawlers.
     def show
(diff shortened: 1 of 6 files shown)


app/controllers/v1/purls_controller.rb
```diff
@@ -14,25 +14,30 @@ def show
     # Update the database purl record from the passed in cocina.
     # This is a legacy API and will be replaced by ResourcesController#create
     def update
-      @purl = begin
-                Purl.find_or_create_by(druid: druid_param)
-              rescue ActiveRecord::RecordNotUnique
-                retry
-              end
-
+      find_purl
```
justinlittman, line 17: Put in before_action ?

justinlittman (approved):
Approve pending consideration of comment.

app/controllers/v1/purls_controller.rb
jcoyne: Sure. Done.

PR #813, opened 2024-06-25
Return the files and their digests for the object

### LINKED ISSUE #811: Return filenames / fixities for the API that DSA calls
So that we don't have to re-write the cocina when the cocina versions used by DSA and persisted in /purl don't match.

The key of the response will be the digest hash (use md5 for now) because there can be multiple versions with the same file name.  Also make it clear that it is a MD5 digest, in case we decide to change this later.

jcoyne:
This will allow DSA to query which files we already have

Fixes #811

### DIFF
--- a/app/models/public_json.rb
+++ b/app/models/public_json.rb
@@ -12,4 +12,15 @@ class PublicJson < ApplicationRecord
   def cocina_hash
     JSON.parse(data)
   end
+
+  # The purpose here is to tell DSA which files we currently have (across all versions)
+  def files_by_md5
+    file_sets = cocina_hash.dig('structural', 'contains')
+    file_sets.flat_map do |fs|
+      fs.dig('structural', 'contains').map do |file|
+        md5_node = file.fetch('hasMessageDigests').find { |digest| digest.fetch('type') == 'md5' }
+        { md5_node.fetch('digest') => file.fetch('filename') }
+      end
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


app/models/public_json.rb
```diff
@@ -8,4 +8,19 @@ def data=(data)
...
     ActiveSupport::Gzip.decompress(super)
   end
+
+  def cocina_hash
+    data && JSON.parse(data)
+  end
+
+  # The purpose here is to tell DSA which files we currently have (across all versions)
+  def files_by_md5
+    file_sets = cocina_hash.dig('structural', 'contains')
```
mjgiarlo, line 18: It's not clear what this method returns if the `JSON.parse(data)` branch is not reached due to `data` being false-y. But in that case, does the thing returned respond to `#dig` (line 18)?

jcoyne: I don't think this ever happens:
```
Purl.where(public_json: nil).count
 => 0
 ```

mjgiarlo: If it's never `nil`, do we need the `data && ` part on line 13? Or are you saying that when that is false-y, `#files_by_md5` is never invoked?

jcoyne: We no longer have that thanks to #816

mjgiarlo: Excellent 🍻

PR #816, opened 2024-06-27
Move JSON deserialization to the PublicJson class

jcoyne:
Remove unnecessary nil checks. We do not have any nil JSON anymore:
```
Purl.where(public_json: nil).count
 => 0

PublicJson.where(data: nil).count
 => 0
```

### DIFF
--- a/app/models/public_json.rb
+++ b/app/models/public_json.rb
@@ -8,4 +8,8 @@ class PublicJson < ApplicationRecord
   def data
     ActiveSupport::Gzip.decompress(super)
   end
+
+  def cocina_hash
+    JSON.parse(data)
+  end
 end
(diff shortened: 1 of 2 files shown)


app/models/public_json.rb
```diff
@@ -8,4 +8,8 @@ def data=(data)
   def data
     ActiveSupport::Gzip.decompress(super)
   end
+
+  def cocina_hash
+    JSON.parse(data)
```
justinlittman, line 13: Memoize?

justinlittman (approved):
Approve pending consideration of comment.

app/models/public_json.rb
jcoyne: Unnecessary. The caller is memoizing the result.

PR #830, opened 2024-07-02
Prevent nil pointer error on already deleted object

### LINKED ISSUE #829: [purl-fetcher] NoMethodError: undefined method `cocina_hash' for nil
## Backtrace

    line 60 of [PROJECT_ROOT]/app/models/purl.rb: cocina_object
    line 26 of [PROJECT_ROOT]/app/controllers/v1/purls_controller.rb: destroy

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/48916/faults/109040535)

jcoyne:
Fixes #829

### DIFF
--- a/app/controllers/v1/purls_controller.rb
+++ b/app/controllers/v1/purls_controller.rb
@@ -23,6 +23,8 @@ module V1
     end
 
     def destroy
+      return render json: { error: "already deleted" }, status: :conflict if @purl.deleted?
+
       @purl.mark_deleted
       UpdateStacksFilesService.delete!(@purl.cocina_object)
       Racecar.produce_sync(value: nil, key: druid_param, topic: Settings.indexer_topic)
--- a/spec/factories/purl.rb
+++ b/spec/factories/purl.rb
@@ -16,5 +16,11 @@ FactoryBot.define do
     trait :with_collection do
       collections { [association(:collection)] }
     end
+
+    trait :deleted do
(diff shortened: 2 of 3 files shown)


app/controllers/v1/purls_controller.rb
```diff
@@ -23,6 +23,8 @@ def update
     end
 
     def destroy
+      return render json: { error: "already deleted" }, status: :conflict if @purl.deleted?
```
mjgiarlo, line 26: Can you say a little more about using HTTP 409 to respond to a request to delete an already deleted resource? I could see 409s being the best choice, but could also see a case being made for 204 or 404. Is it helpful to consumers to return a 4xx status? Does that signal "do no more, the deed is done?"

mjgiarlo (approved):
Approve pending consideration of HTTP status question.

app/controllers/v1/purls_controller.rb
jcoyne: Yes, I think it's the choice because you can't make a transition from the current state.  It's not an acceptable request.  This is what we use in DSA on similar occurrences: https://github.com/sul-dlss/dor-services-app/blob/fffc87e50b2716f1d36b39876ded9abf834d0cac/openapi.yml#L159-L160

mjgiarlo: Sounds good, thanks

PR #844, opened 2024-07-10
Remove ActiveStorage file transfer in favor of a transfer directory

jcoyne:


app/services/update_stacks_files_service.rb
```diff
@@ -34,31 +33,28 @@ def delete!
...
     raise RequestError, 'Files in file_uploads not in cocina object'
   end
 
-  def check_signed_ids
-    return if file_uploads_map.values.all? { |signed_id| ActiveStorage.verifier.valid_message?(signed_id) }
-
-    raise RequestError, "Invalid signed ids found"
-  end
-
   # Copy the files from ActiveStorage to the Stacks directory
```
justinlittman, line 46: Comment needs update.

app/services/update_stacks_files_service.rb
```diff
@@ -34,31 +33,28 @@ def delete!
...
   def shelve_files
     file_uploads_map.each do |filename, signed_id|
```
justinlittman, line 48: Can `signed_id` be renamed?

justinlittman (approved):
Approved pending consideration of comments.

app/services/update_stacks_files_service.rb
jcoyne: done

app/services/update_stacks_files_service.rb
jcoyne: done

PR #853, opened 2024-07-15
Encapsulate awfl.

justinlittman:
* Adds an AWFL service for creating / updating / deleting versions and migrating from unversioned to versioned layout.
* Updates the controllers to use AWFL based on configuration, whether the object is already versioned, and whether the object is new.
* Removes the previous, initial AWFL implementation.


app/controllers/v1/purls_controller.rb
```diff
@@ -17,8 +17,19 @@ def destroy
...
       @purl.mark_deleted
-      UpdatePurlMetadataService.new(@purl).delete!
-      UpdateStacksFilesService.delete!(@purl.cocina_object)
+
+      if Settings.features.versioned_files && VersionedFilesService.versioned_files?(druid: @purl.druid)
+        # TODO: Allow DSA to provide the version to delete. Hardcoding version to 1 for now.
+        begin
+          VersionedFilesService.new(druid: @purl.druid).delete(version: '1')
+        rescue VersionedFilesService::UnknowVersionError
+          # This shouldn't happen, but in case it does it can be ignored.
```
jcoyne, line 26: Can you expand on why it shouldn't  or in which cases it would?

app/controllers/v1/purls_controller.rb
```diff
@@ -17,8 +17,19 @@ def destroy
       return render json: { error: "already deleted" }, status: :conflict if @purl.deleted?
 
       @purl.mark_deleted
-      UpdatePurlMetadataService.new(@purl).delete!
-      UpdateStacksFilesService.delete!(@purl.cocina_object)
+
+      if Settings.features.versioned_files && VersionedFilesService.versioned_files?(druid: @purl.druid)
```
jcoyne, line 21: Can this conditional go in the `UpdateStacksFilesService`?

app/controllers/v1/resources_controller.rb
```diff
@@ -12,16 +12,33 @@ class ResourcesController < ApplicationController
...
+      #   invoke VersionedFilesService.migrate before VersionedFilesService.update to migrate from unversioned to versioned layout.
+      if versioned?(druid: @purl.druid)
+        # Writes to stacks with versioned layout.
+        VersionedFilesService.new(druid: @purl.druid).update(version: '1',
+                                                             version_metadata: VersionedFilesService::VersionMetadata.new(withdrawn: false, date: DateTime.now),
+                                                             cocina: @cocina_object,
+                                                             public_xml: PublicXmlWriter.generate(@cocina_object),
+                                                             file_transfers: file_uploads)
+        # Writes to purl. In the future when PURL Application can handle versioned layout, this will be removed.
+        UpdatePurlMetadataService.new(@purl).write! if Settings.features.legacy_purl
```
jcoyne, line 35: This is in both blocks of the if/else.  Can we move it outside?

app/controllers/v1/resources_controller.rb
```diff
@@ -12,16 +12,33 @@ class ResourcesController < ApplicationController
...
     def create
       PurlCocinaUpdater.new(@purl, @cocina_object).update
 
-      UpdateStacksFilesService.write!(@cocina_object, file_uploads) unless @cocina_object.collection?
-      UpdatePurlMetadataService.new(@purl).write!
+      # TODO: Allow DSA to provide a version. Hardcoding version to 1 for now
+      # TODO: Allow DSA to provide a version date. Setting version date to now for now.
+      # TODO: Allow DSA to indicate if an object is versioned. If versioned and object is not using versioned layout,
+      #   invoke VersionedFilesService.migrate before VersionedFilesService.update to migrate from unversioned to versioned layout.
+      if versioned?(druid: @purl.druid)
```
jcoyne, line 27: Can this all be moved into the UpdateStacksFilesService?

app/services/versioned_files_service/paths.rb
```diff
@@ -0,0 +1,88 @@
+class VersionedFilesService
+  # Support for paths.
+  class Paths
```
jcoyne, line 3: I don't like how this mixes a bunch of different kinds of paths.  I think the AwflPaths could be in on class, but the transfer path and purl paths should be in separate classes.

app/services/versioned_files_service/link_support.rb
```diff
@@ -0,0 +1,23 @@
...
+  class LinkSupport
+    # @param [Pathname] source_path
+    # @param [Pathname] dest_path
+    def self.link(source_path, dest_path)
+      # There should be no need for this check. However we're not seeing the file on the filesystem, so check for now.
+      raise "Path doesn't exist: `#{source_path}'" unless source_path.exist?
+
+      return if link?(source_path, dest_path)
+
+      dest_path.delete if dest_path.exist? || dest_path.symlink?
```
jcoyne, line 11: Why are we checking for symlinks here?

app/services/versioned_files_service/update_action.rb
```diff
@@ -0,0 +1,104 @@
...
+        delete_content(md5:)
+      end
+    end
+
+    def cocina_content_md5s
+      versions.map do |version|
+        cocina_path = cocina_path_for(version:)
+        raise VersionedFilesService::Error, "Cocina for version #{version} not found" unless cocina_path.exist?
+
+        cocina = Cocina.new(hash: JSON.parse(cocina_path.read))
```
jcoyne, line 99: Can this be a method in the Cocina class (e.g. `Cocina.for_version(version)`)?

app/services/versioned_files_service/migrate_action.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def link_content_files
+      FileUtils.mkdir_p(content_path)
+      shelve_file_map.each do |filename, md5|
+        LinkSupport.link(stacks_content_path_for(filename:), content_path_for(md5:))
+      end
+    end
+
+    def shelve_file_map
+      @shelve_file_map ||= Cocina.new(hash: cocina_hash).shelve_file_map
```
jcoyne, line 43: Can this be a method in the `Cocina` class (e.g. `Cocina.for_version(version:)`)

app/services/versioned_files_service/purge_content_action.rb
```diff
@@ -0,0 +1,34 @@
...
+        delete_content(md5:)
+      end
+    end
+
+    def cocina_content_md5s
+      versions.map do |version|
+        cocina_path = cocina_path_for(version:)
+        raise VersionedFilesService::Error, "Cocina for version #{version} not found" unless cocina_path.exist?
+
+        cocina = Cocina.new(hash: JSON.parse(cocina_path.read))
```
jcoyne, line 29: Can this be a method in the Cocina class (e.g. `Cocina.for_version(version)`)?

app/services/versioned_files_service/update_action.rb
```diff
@@ -0,0 +1,104 @@
...
+      file_transfers.each do |filename, transfer_uuid|
+        md5 = shelve_file_map[filename]
+        next if content_md5s.include?(md5)
+
+        move_content(source_path: transfer_path_for(transfer_uuid:), md5:)
+      end
+    end
+
+    def shelve_file_map
+      @shelve_file_map ||= Cocina.new(hash: cocina.to_h).shelve_file_map
```
jcoyne, line 77: Can this be a method in the Cocina class (e.g. `Cocina.for_version(version)`)?

config/settings.yml
```diff
@@ -14,12 +14,10 @@ hmac_secret: ~
 filesystems:
   purl_root: "/purl/document_cache"
   stacks_root: "/stacks"
-  stacks_content_addressable: /stacks/content_addressable
```
jcoyne, line 17: Are we not using this path for versioned objects anymore?  I thought we agreed (with Julian) that we would put all of the versioned objects in a separate root.

app/services/versioned_files_service/contents.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      content_path.children.map { |child| child.basename.to_s }
+    end
+
+    def move_content(md5:, source_path:)
+      FileUtils.mkdir_p(content_path)
+      FileUtils.mv(source_path, content_path_for(md5:))
+    end
+
+    def delete_content(md5:)
```
jcoyne, line 20: Can this move to `PurgeContentAction`? It's only used there and we have to delegate to it a bunch of times.

app/services/versioned_files_service/migrate_action.rb
```diff
@@ -0,0 +1,58 @@
+class VersionedFilesService
+  # Migrate an object from unversioned to versioned Stacks layout.
+  class MigrateAction
+    def initialize(service:, version_metadata:)
```
jcoyne, line 4: What are these parameters? Can you document them?

app/services/versioned_files_service/metadata.rb
```diff
@@ -0,0 +1,45 @@
+class VersionedFilesService
+  # Support for managing cocina.json and public xml files.
+  class Metadata
+    def initialize(service:)
```
jcoyne, line 4: What are this parameter? Can you document?

app/services/versioned_files_service/contents.rb
```diff
@@ -0,0 +1,26 @@
+class VersionedFilesService
+  # Support for managing content files.
+  class Contents
+    def initialize(service:)
```
jcoyne, line 4: What are this parameter? Can you document?

app/services/versioned_files_service/migrate_action.rb
justinlittman: Well, took me trying to implement to realize this, but: This is loading the cocina from the PURL file system. There is no version involved and I think this is the only place where we would want to do this.

app/services/versioned_files_service/paths.rb
justinlittman: Agree. Done.

app/services/versioned_files_service/purge_content_action.rb
justinlittman: Yes. Done.

app/services/versioned_files_service/update_action.rb
justinlittman: No, because this is generated from the provided Cocina::Models::*, not based on a versioned cocina.json on disk.

app/services/versioned_files_service/update_action.rb
justinlittman: Yes. Done.

config/settings.yml
justinlittman: I proposed putting all of the versioned objects in a separate root. I don't think Julian was a hard no on it, but preferred that it not be in a separate root (so that the linked files would all be close together and definitely on the same file system). I didn't see any compelling reason not to take that approach.

app/controllers/v1/purls_controller.rb
jcoyne: I think it's important to encapsulate logic like "where are we storing this stuff", and that encapsulation should protect the controller from needing to know this.  I think it's possible to put this in UpdateStacksFilesService (even if you need to move what is currently there to "LegacyStacksFilesService".

justinlittman: OK @jcoyne ; this is done.

jcoyne (approved):

PR #861, opened 2024-07-22
Adds missing method documentation.

justinlittman:


app/services/purl_and_stacks_service.rb
```diff
@@ -7,13 +8,16 @@ def self.update(purl:, cocina_object:, file_uploads:, version:, version_date:)
...
 
+  # @param purl [Purl] the PURL model object.
   def initialize(purl:)
     @purl = purl
   end
 
+  # Update the PURL and Stacks files.
+  # @param version [String] the version number
+  # @param version_date [DateTime] the version date
   def update(cocina_object:, file_uploads:, version:, version_date:)
```
mjgiarlo, line 19: Are the `cocina_object` and `file_uploads` params intentionally left out of the documentation?

justinlittman: I don't know where they went; there now.

mjgiarlo (approved):

app/services/purl_and_stacks_service.rb
mjgiarlo: Who can blame them for running away?

Issue #862, opened 2024-07-22
Add additional API parameters for versioning


justinlittman:
For creating a PURL:
* version
* version_date
* must_version

For deleting a PURL:
* version

justinlittman:
Done.

PR #863, opened 2024-07-22
Adds additional API params for updating and deleting versions.

justinlittman:
refs #862


app/services/versioned_files_service.rb
```diff
@@ -60,7 +60,7 @@ def delete(version:)
 
   # Migrate from unversioned to versioned layout.
   # @param version_metadata [VersionMetadata] the metadata for the version
-  def migate(version_metadata:)
```
aaron-collier, line 63: Good catch.

aaron-collier (approved):

PR #866, opened 2024-07-23
Move stacks_object_path into `Paths`

cbeer:


justinlittman:
@jcoyne asked me to do the opposite of this.

cbeer:
I'm not sure I understand the distinction being made here. All of them sure look like paths helpers to me.

justinlittman:
The paths in the Paths class are all paths within the versioned layout. (https://github.com/sul-dlss/purl-fetcher/blob/main/app/services/versioned_files_service/paths.rb#L2)

cbeer:
I guess I don't understand why that is a useful distinction. I'd much rather have one place to see all the paths/patterns we're using than having this same-as-but-very-slightly-distinguisable method hanging out somewhere else. I think it just makes the code harder to grok 🤷

PR #867, opened 2024-07-23
Contents + Metadata only needs paths, so we don't need to provide the whole service.

cbeer:
The pattern of sending the whole service to the ~10 supporting actions + classes makes it hard for me to understand what's going on. It looks like some of these only need the paths helper, and maybe that will help my understanding.

### DIFF
--- a/app/services/versioned_files_service.rb
+++ b/app/services/versioned_files_service.rb
@@ -21,12 +21,9 @@ class VersionedFilesService
   def initialize(druid:)
     @druid = druid
     @paths = Paths.new(druid:)
-    @version_manifest = VersionsManifest.new(path: @paths.versions_manifest_path)
-    @contents = Contents.new(service: self)
-    @metadata = Metadata.new(service: self)
   end
 
-  attr_reader :druid, :version_manifest
+  attr_reader :druid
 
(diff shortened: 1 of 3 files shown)


app/services/versioned_files_service.rb
```diff
@@ -21,12 +21,9 @@ def self.versioned_files?(druid:)
   def initialize(druid:)
     @druid = druid
     @paths = Paths.new(druid:)
-    @version_manifest = VersionsManifest.new(path: @paths.versions_manifest_path)
-    @contents = Contents.new(service: self)
-    @metadata = Metadata.new(service: self)
   end
```
mjgiarlo, line 24: Do you think giving line 23 the same treatment is worth considering? (i.e., extracting the ivar into a new instance method that memoizes the same ivar)

mjgiarlo: Oh, nevermind. I now see the other extracted methods depend on `@paths` already.

cbeer: I thought about it, but would rather do that in a separate PR if we think it's a good idea.

justinlittman:
Approve pending resolving the conflict.

PR #870, opened 2024-07-23
Always use a versioned layout if the feature flag is set.

cbeer:
I'm not sure I understand how we're using the `must_version` flag (or when it would or would not be set).

Could we just always use the versioned layout if we're touching an object (because the believe the new layout is now backwards compatible?)


cbeer:
Andrew periodically republishes object for administrative reasons. We want to avoid migrating those objects (for now?).

PR #872, opened 2024-07-23
Update DublinCoreService to take the cocina and generate needed assets internally

### LINKED ISSUE #858: DOI link is shown twice
The DOI link at the bottom of the purl page is being repeated:

Screenshot 2024-07-19 at 1 32 05 PM

It looks like this is a recent problem. I've noticed that if I republish an item whose purl shows only one DOI link, the purl page will show two DOI links after republishing.

- example with two DOI links: https://purl.stanford.edu/cp231wr9194

- example with one DOI link: https://purl.stanford.edu/qz834fk6040 (republishing the item will likely make it show two DOI links)

cbeer:
This fixes #858 by making sure we're not accidentally sharing state.

Alternative to #860 .

### DIFF
--- a/app/services/publish/public_desc_metadata_service.rb
+++ b/app/services/publish/public_desc_metadata_service.rb
@@ -3,16 +3,16 @@
 module Publish
   # Creates the descriptive XML that we display on purl.stanford.edu
   class PublicDescMetadataService
-    attr_reader :cocina_object, :constituents
+    attr_reader :cocina_object, :constituents, :include_access_conditions
 
     MODS_NS = 'http://www.loc.gov/mods/v3'
 
     # @param [Cocina::Models::Collection,Cocina::Models::DRO] cocina_object
     # @param [Array<Hash>] constituents a list of constituents (virtual object members) that are part of this object
-    def initialize(cocina_object, constituents)
+    def initialize(cocina_object, constituents, include_access_conditions: true)
       @cocina_object = cocina_object
       @constituents = constituents
-      @ng_xml = {}
+      @include_access_conditions = include_access_conditions
     end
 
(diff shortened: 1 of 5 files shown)


app/services/publish/public_desc_metadata_service.rb
```diff
@@ -3,16 +3,17 @@
...
   class PublicDescMetadataService
-    attr_reader :cocina_object, :constituents
+    attr_reader :cocina_object, :constituents, :include_access_conditions
 
     MODS_NS = 'http://www.loc.gov/mods/v3'
 
     # @param [Cocina::Models::Collection,Cocina::Models::DRO] cocina_object
     # @param [Array<Hash>] constituents a list of constituents (virtual object members) that are part of this object
-    def initialize(cocina_object, constituents)
+    def initialize(cocina_object, constituents, include_access_conditions: true)
```
justinlittman, line 12: Can you please add the doc for the new parameter?

app/services/publish/public_desc_metadata_service.rb
```diff
@@ -3,16 +3,17 @@
...
       @cocina_object = cocina_object
       @constituents = constituents
-      @ng_xml = {}
+      @include_access_conditions = include_access_conditions
+      @ng_xml = nil
```
justinlittman, line 16: Is L16 necessary at all?

justinlittman (approved):
Approve pending consideration of comments.

app/services/publish/public_desc_metadata_service.rb
cbeer: I would love to, if only I knew what it meant or why it was used. The `#ng_xml` method it came from didn't document that information.

PR #877, opened 2024-07-23
Use a distinct parameter name for released#index.

cbeer:
It's a little disconcerting if `show` uses `id` and update uses `druid`...

### DIFF
--- a/app/controllers/v1/released_controller.rb
+++ b/app/controllers/v1/released_controller.rb
@@ -7,7 +7,7 @@ module V1
     # Query for all druids that have a certain release tag.
     # This is used by the sitemap generator in PURL.
     def show
-      release_tag = params[:id]
+      release_tag = params[:release_tag]
       purls = Purl.published
                   .status('public')
                   .target(release_tag).pluck(:druid, :updated_at)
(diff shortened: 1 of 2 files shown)


app/controllers/v1/released_controller.rb
```diff
@@ -7,7 +7,7 @@ class ReleasedController < ApplicationController
     # Query for all druids that have a certain release tag.
     # This is used by the sitemap generator in PURL.
     def show
-      release_tag = params[:id]
+      release_tag = params[:druid]
```
justinlittman, line 10: This is a release tag, not a druid.

justinlittman (changes requested):

app/controllers/v1/released_controller.rb
cbeer: Thanks, I didn't catch that at all. Isn't it pretty weird to have `GET /released/*` and `PUT /released/*` do totally different things?

justinlittman: ^^ cc @jcoyne

justinlittman (approved):

PR #892, opened 2024-07-25
Refactor VersionsManifest to include a schema version attribute

cbeer:
... and use the model objects for testing the serialization.

### DIFF
--- a/app/services/versioned_files_service/versions_manifest.rb
+++ b/app/services/versioned_files_service/versions_manifest.rb
@@ -86,7 +86,10 @@ class VersionedFilesService
 
     def manifest
       @manifest ||= (path.exist? ? JSON.parse(@path.read).with_indifferent_access : {}).tap do |manifest|
+        manifest[:$schemaVersion] ||= 1
+
         # json numeric keys are converted to strings, so convert them back to integers
+        manifest[:versions] ||= {}
         manifest[:versions]&.transform_keys!(&:to_i)
         manifest[:head] &&= manifest[:head].to_i
       end
--- a/app/services/versioned_files_service/versions_manifest.rb
+++ b/app/services/versioned_files_service/versions_manifest.rb
@@ -1,6 +1,10 @@
 class VersionedFilesService
   # Class for reading and writing the versions manifest.
   class VersionsManifest
+    def self.read(path)
+      new(path: Pathname.new(path))
+    end
+
     # @param path [Pathname] the path to the versions manifest
     def initialize(path:)
       @path = path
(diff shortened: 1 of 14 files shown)


app/services/versioned_files_service/versions_manifest.rb
```diff
@@ -77,24 +92,23 @@ def version_metadata
...
+      manifest[:versions].keys
     end
 
-    private
-
-    attr_reader :path
-
     def manifest
       @manifest ||= (path.exist? ? JSON.parse(@path.read).with_indifferent_access : {}).tap do |manifest|
+        manifest[:@version] ||= 1
```
justinlittman, line 100: Calling this version is likely to cause confusion.

spec/services/versioned_files_service/object_spec.rb
```diff
@@ -40,8 +40,8 @@
     end
 
     context 'when the version manifest does not exist' do
-      it 'raises UnknownVersionError' do
-        expect { service.head_version }.to raise_error(VersionedFilesService::UnknowVersionError, 'Head version not found')
+      it 'returns 0' do
```
justinlittman, line 43: Returning 0 doesn't seem like a clear way of indicating that there is no version.

cbeer: It's not clear to me that purl-fetcher needs to care whether it had an already published version or not. It seems like it just needs to know whether the version being offered is newer than what's already there.

app/services/versioned_files_service/versions_manifest.rb
cbeer: Would you be happier with `"$schema": "....#v1"`? That struck me as super over-kill, but maybe it's less confusing

spec/services/versioned_files_service/object_spec.rb
cbeer: I've pushed up `090752a` that uses `nil` instead of `0`. I'm not sure having to scatter nil-checks around is a great improvement 🤷‍♂️

justinlittman: Why not leave it as raising an UnknownVersionError?

cbeer: I guess not having a head version doesn't seem very exceptional (or particularly problematic) to me.

I think having `nil` or some nil-like value is much clearer than a method that throws an exception and having to remember to also check `head_version?` every time you want to access `head_version` (so you can avoid throwing an exception).

justinlittman (approved):

PR #897, opened 2024-07-25
Refactor ReleaseService

cbeer:


app/services/release_service.rb
```diff
@@ -2,30 +2,44 @@ class ReleaseService
...
     file_path = File.join(purl.purl_druid_path, 'meta.json')
-    File.write(file_path, meta_json(purl))
+    File.write(file_path, meta_json)
   end
-  private_class_method :write_purl_meta_json
 
-  def self.meta_json(purl)
+  def meta_json
     {
+      '$schemaVersion': 1,
```
justinlittman, line 42: Being able to make this change without changing a test seems to indicate missing tests.

cbeer: That's one of the reasons I thought this class would benefit from a little refactoring. `private_class_method` seemed like a code-smell.

justinlittman (approved):

PR #898, opened 2024-07-26
Add an action to withdraw a version

### LINKED ISSUE #875: API support for withdrawing a version
This will result in the appropriate update in the object's version manifest.

refs https://github.com/sul-dlss/argo/issues/4528

cbeer:
Fixes #875

### DIFF
--- a/app/controllers/v1/purls_controller.rb
+++ b/app/controllers/v1/purls_controller.rb
@@ -47,6 +47,12 @@ module V1
       render json: true, status: :accepted
     end
 
+    def withdraw_version
+      return render json: { error: "not yet published" }, status: :conflict if @purl.new_record?
+
+      PurlAndStacksService.withdraw(purl: @purl, version:)
+    end
+
     private
 
     def find_purl
--- a/app/services/purl_and_stacks_service.rb
+++ b/app/services/purl_and_stacks_service.rb
@@ -20,7 +20,7 @@ class PurlAndStacksService
   # @param version_date [DateTime] the version date
   # @param must_version [Boolean] true if the versioned layout is required
   def update(cocina_object:, file_uploads:, version:, version_date:, must_version:)
-    version_metadata = VersionedFilesService::VersionMetadata.new(version: version.to_i, withdrawn: false, date: version_date)
+    version_metadata = VersionedFilesService::VersionsManifest::VersionMetadata.new(version: version.to_i, withdrawn: false, date: version_date)
     VersionedFilesService.new(druid:).migrate(version_metadata:) if migrate_to_versioned_layout?(must_version)
 
     if use_versioned_layout?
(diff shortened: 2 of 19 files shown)


app/services/purl_and_stacks_service.rb
```diff
@@ -56,6 +56,18 @@ def delete(version:)
     UpdatePurlMetadataService.new(purl).delete! if legacy_purl_enabled?
   end
 
+  # Delete the PURL and Stacks files.
+  # @param version [String] the version number
+  def withdraw(version:)
+    return unless versioned_files_enabled? && already_versioned_layout?
```
justinlittman, line 62: I think this should raise. Trying to withdraw an unversioned object indicates that something is wrong.

app/controllers/v1/purls_controller.rb
```diff
@@ -47,6 +47,12 @@ def release_tags
       render json: true, status: :accepted
     end
 
+    def withdraw_version
+      return render json: { error: "not yet published" }, status: :conflict if @purl.new_record?
```
justinlittman, line 51: Probably should also raise if purl status is deleted.

app/services/versioned_files_service/withdraw_action.rb
```diff
@@ -0,0 +1,33 @@
+class VersionedFilesService
+  # Deletes head version.
```
justinlittman, line 2: Comment is wrong.

app/services/versioned_files_service/withdraw_action.rb
```diff
@@ -0,0 +1,33 @@
...
+  class WithdrawAction
+    # @param version [String] the version number
+    # @param object [VersionedFilesService::Object] the object
+    def initialize(version:, object:)
+      @version = version
+      @object = object
+    end
+
+    # @raise [UnknowVersionError] if the version is not found
+    # @raise [Error] if the version is not the head version
```
justinlittman, line 12: if the version is the head version.

app/services/versioned_files_service/withdraw_action.rb
```diff
@@ -0,0 +1,33 @@
...
+    def call
+      raise VersionedFilesService::UnknowVersionError, "Version #{version} not found" unless version?(version:)
+      raise VersionedFilesService::Error, "The head version cannot be withdrawn, only be deleted" if head_version == version
+
+      # Update the version manifest.
+      version_manifest.withdraw(version:)
+      # Delete the cocina version file and set the new head symlink.
```
justinlittman, line 19: Cocina, public xml, and content files shouldn't be changed. (Just the version manifest.)

justinlittman (commented):
Also going to need to restore (that is, undo a withdraw).

PR #899, opened 2024-07-26
Add object locking around updates + deletes

cbeer:
Or maybe there's something else in the architecture that's stopping simultaneous updates?

### DIFF
--- a/app/services/versioned_files_service/migrate_action.rb
+++ b/app/services/versioned_files_service/migrate_action.rb
@@ -11,23 +11,32 @@ class VersionedFilesService
     def call
       # For each shelved file in the cocina object, make sure there is a content file.
       check_content_files!
-      # For each shelved file, create a hardlink named by md5
-      link_content_files
-      # Write the cocina to cocina path for the version and create a new head cocina symlink.
-      write_cocina(version: 1, cocina: cocina_hash, head_version: true)
-      # Write the public xml to public xml path for the version and create a new head public xml symlink .
-      write_public_xml(version: 1, public_xml:, head_version: true)
-      copy_meta_json
-      # Update the version manifest.
-      version_manifest.update_version(version: 1, version_metadata:)
+
+      create_directory_structure
+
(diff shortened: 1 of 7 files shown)


justinlittman:
Will second attempt to lock block or raise? (I don't love blocking, as it might lock up http requests. I'd prefer to see it raise so that the caller can retry.)

cbeer:
We can make it do whatever. It's not yet clear to me whether this is needed (but also not clear that it isn't.). What do you think?

justinlittman:
In theory it shouldn't happen. However, since on the DSA side publishing is performed by a job, anything is possible given retries. This seems lightweight enough that it is worth it.

app/services/versioned_files_service/migrate_action.rb
```diff
@@ -9,26 +9,34 @@ def initialize(object:, version_metadata:)
...
-      # For each shelved file, create a hardlink named by md5
-      link_content_files
-      # Write the cocina to cocina path for the version and create a new head cocina symlink.
-      write_cocina(version: 1, cocina: cocina_hash, head_version: true)
-      # Write the public xml to public xml path for the version and create a new head public xml symlink .
-      write_public_xml(version: 1, public_xml:, head_version: true)
-      copy_meta_json
-      # Update the version manifest.
-      version_manifest.update_version(version: 1, version_metadata:, head_version: true)
+      create_directory_structure
```
justinlittman, line 12: I don't think we should create the versions directory before checking the content files. Once the versions directory is created, applications will think that the object is using the versioned layout, even though the directory is empty.

justinlittman: If you just switch `check_content_files!` and `create_directory_structure` I think the logic should be fine.

cbeer: Except the lock file is created within the AWFL directory structure?

justinlittman (approved):

PR #902, opened 2024-08-01
Support withdrawing and restoring versions.

### LINKED ISSUE #875: API support for withdrawing a version

justinlittman:
closes #875

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,10 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery unless: -> { request.format.json? }, with: :exception
 
+  rescue_from VersionedFilesService::Lock::LockError do |e|
+    render build_error('423', e, 'Error acquiring lock')
+  end
+
   private
 
   def page_params
@@ -17,5 +21,22 @@ class ApplicationController < ActionController::Base
     druid = params[:druid]
     druid = "druid:#{druid}" unless druid&.start_with?('druid:')
     druid
+  end
+
+  # JSON-API error response. See https://jsonapi.org/.
+  def build_error(error_code, err, msg)
+    {
+      json: {
+        errors: [
+          {
+            status: error_code,
+            title: msg,
+            detail: err.message
+          }
+        ]
+      },
+      content_type: 'application/json',
+      status: error_code
+    }
   end
 end
--- a/app/controllers/v1/versions_controller.rb
+++ b/app/controllers/v1/versions_controller.rb
@@ -1,0 +1,24 @@
+module V1
+  class VersionsController < ApplicationController
+    include Authenticated
+
+    before_action :check_auth_token
+    before_action :find_purl
+
+    rescue_from VersionedFilesService::BadRequestError do |e|
+      render build_error('400', e, 'Bad request')
+    end
+
+    def withdraw
+      return render json: { error: "already deleted" }, status: :conflict if @purl.deleted?
+
+      PurlAndStacksService.withdraw(purl: @purl, version: params[:version], withdrawn: params[:withdrawn])
+    end
+
+    private
+
+    def find_purl
+      @purl = Purl.find_by!(druid: druid_param)
+    end
+  end
+end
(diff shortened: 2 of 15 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,7 +1,11 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery unless: -> { request.format.json? }, with: :exception
 
-  private
+  rescue_from VersionedFilesService::Lock::LockError do |e|
+    render build_error('423', e, 'Error acquiring lock')
+  end
+
+  protected
```
mjgiarlo, line 8: Can you say more about choosing `protected` over `private` here?

config/routes.rb
```diff
@@ -29,6 +29,13 @@
 
         put 'release_tags'
         patch 'release_tags'
+
+        resources :versions, only: [], param: :version do
+          member do
+            match 'withdraw', via: %i[put patch], defaults: { withdrawn: true }, to: 'versions#withdraw'
+            match 'restore', via: %i[put patch], defaults: { withdrawn: false }, to: 'versions#withdraw'
+          end
+        end
```
mjgiarlo, line 38: Is it worth considering a resource-oriented approach here, where PUTs to `/v1/purls/:druid/versions/:version` will restore and DELETEs to the same URL will withdraw?

app/controllers/application_controller.rb
justinlittman: Adjusting so that not necessary.

app/controllers/v1/versions_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+    before_action :find_purl
+
+    rescue_from VersionedFilesService::BadRequestError do |e|
+      render build_error('400', e, 'Bad request')
+    end
+
+    def withdraw
+      return render json: { error: "already deleted" }, status: :conflict if @purl.deleted?
+
+      PurlAndStacksService.withdraw(purl: @purl, version: params[:version], withdrawn: params[:withdrawn])
```
mjgiarlo, line 15: Not seeing a render when this succeeds. Is the default controller behavior here to return an empty 204 response?

config/routes.rb
justinlittman: It isn't really a delete, in that the resource isn't destroyed, it's visibility is just changed. In addition, restoring isn't the only update that one might want to make to a version.

I'm not strongly opposed to the change, but not convinced it is a great fit.

app/services/purl_and_stacks_service.rb
```diff
@@ -56,6 +60,16 @@ def delete(version:)
...
 
+  # Withdraw or restore a version.
+  # @param version [String] the version number
+  # @param withdrawn [Boolean] true to withdraw, false to restore
+  def withdraw(version:, withdrawn: true)
+    # unversioned is an implicit version 1, which cannot be withdrawn.
+    raise VersionedFilesService::BadRequestError, 'Cannot withdraw head version' unless already_versioned_layout?
+
+    versioned_files_service.withdraw(version:, withdrawn:)
+  end
```
mjgiarlo, line 71: I guess the logic hits the versioned files service via this service because this is the class responsible for determining if the layout is versioned?

config/application.rb
```diff
@@ -37,5 +37,6 @@ class Application < Rails::Application
     #
     # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
+    config.action_dispatch.rescue_responses['VersionedFilesService::UnknowVersionError'] = :not_found
```
mjgiarlo, line 40: Typo?

```suggestion
    config.action_dispatch.rescue_responses['VersionedFilesService::UnknownVersionError'] = :not_found
```

config/routes.rb
mjgiarlo: Let's hold off on that change for now, then

app/services/purl_and_stacks_service.rb
justinlittman: PurlAndStacksService has been serving as the level of abstraction that is aware of version layout.

mjgiarlo: 👍🏻 Makes sense to isolate that so it's not leaked all over.

app/controllers/v1/versions_controller.rb
justinlittman: Yes.

mjgiarlo (approved):

PR #908, opened 2024-08-07
Backfill missing versions in versions manifest.

### LINKED ISSUE #907: Version manifest should recreate missing versions
When writing a new object and the version > 1, entries should be made in the version manifest for previous versions. Those previous versions should be marked as withdrawn and have no date. The head version should be set to the version.

justinlittman:
closes #907


jcoyne (changes requested):
I don't think this is necessary and it may actually complicate PURL as it has to understand nil dates now.

jcoyne:
@justinlittman i've reversed myself from earlier. Since we want to show these "permanently withdrawn" versions on the purl page, we will need this.

justinlittman:
@jcoyne Fixed up and ready for review.

spec/services/versioned_files_service/versions_manifest_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  describe '.update_version' do
+    context 'when backfilling withdrawn versions' do
+      let(:now) { DateTime.now }
+
+      let(:expected_manifest) do
+        {
+          "$schemaVersion" => 1,
+          'head' => 3,
+          'versions' => {
+            '1' => { 'state' => 'permanently_withdrawn' },
```
jcoyne, line 23: Purl is expecting no underscore https://github.com/sul-dlss/purl/blob/1a60b041a89359b61ee9b9c0133ee2b5a983342f/spec/fixtures/document_cache/wp/335/yr/5649/wp335yr5649/versions/versions.json#L3

jcoyne (changes requested):

spec/services/versioned_files_service/versions_manifest_spec.rb
justinlittman: Can we change purl? DSA also uses permanently_withdrawn and it would be nice to be consistent across systems.

justinlittman: Also, the original ticket specifies `permanently_withdrawn`: https://github.com/sul-dlss/purl/issues/1101

PR #911, opened 2024-08-07
[HOLD] Delete all files through VersionFileService.delete

aaron-collier:
Fixes #906 

TODO: Update to allow delete of withdrawn versions.

I'm not sure this is the appropriate direction to take this, but happy to huddle if warranted.

PR #912, opened 2024-08-08
Extract DarkenService from UpdateStacksFileService for migration to v…

aaron-collier:
…ersioned objects

Follow up from #911

### DIFF
--- a/app/services/darken_service.rb
+++ b/app/services/darken_service.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+# Service for darkening an object.
+# This service removes files from the stacks filesystem in order to make an objct dark.
+class DarkenService
+  def self.call(...)
+    new(...).call
+  end
+
+  def initialize(druid)
+    @stacks_druid_path = DruidTools::PurlDruid.new(druid, Settings.filesystems.stacks_root).pathname
+  end
+
+  # Remove the object from the stacks filesystem and clean up any empty directories in the druid tree.
+  def call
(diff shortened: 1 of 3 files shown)


app/services/darken_service.rb
```diff
@@ -0,0 +1,32 @@
...
+  def self.darken(cocina_object)
+    new(cocina_object).darken
+  end
+
+  def initialize(cocina_object)
+    @stacks_druid_path = DruidTools::PurlDruid.new(cocina_object.externalIdentifier, Settings.filesystems.stacks_root).pathname
+  end
+
+  # Remove the object from the stacks filesystem and clean up any empty directories in the druid tree.
+  def darken
```
justinlittman, line 15: Minor suggestion. For a service that only has a single method, name that method `call`.

app/services/darken_service.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+# Service for darkening an object.
+# This service removes files from the stacks filesystem in order to make an objct dark.
+class DarkenService
+  def self.darken(cocina_object)
+    new(cocina_object).darken
+  end
+
+  def initialize(cocina_object)
```
justinlittman, line 10: Why pass in the cocina object? Looks like all it needs is the druid.

justinlittman (approved):
Approve pending consideration of comments.

aaron-collier:
@justinlittman done. Thanks.

jcoyne (approved):

PR #931, opened 2024-10-08
Logging during publish

edsu:
To help diagnose some timeouts it would be helpful to add a bit of logging around service calls during publish.


edsu:
I guess this isn't needed since we already know where the time is being spent:

From @jcoyne in Slack:

> We know that https://github.com/sul-dlss/purl-fetcher/blob/main/app/services/purl_and_stacks_service.rb#L22-L39 takes a long time on a huge object.

PR #932, opened 2024-10-11
Handle purls missing a public json record.

lwrubel:
Temporarily prevent these items from erroring on republish because they currently are missing public JSON records. 

Tested by modifying a purl on purl-fetcher stage to have no public json and a cocina.json without currently valid cocina. Republished successfully, with no errors, creating a PublicJson record and updating the cocina.json

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -19,5 +19,5 @@ jobs:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
       - run: bundle exec rake
-      - name: Run tests (with content addressed storage feature flag enabled)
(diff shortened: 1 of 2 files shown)


justinlittman (approved):

.github/workflows/ruby.yml
```diff
@@ -19,5 +19,5 @@ jobs:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
       - run: bundle exec rake
-      - name: Run tests (with content addressed storage feature flag enabled)
-        run: SETTINGS__FEATURES__AWFL_METADATA=true SETTINGS__FEATURES__AWFL=true bundle exec rake
+      - name: Run tests
+        run: bundle exec rake
```
jcoyne, line 23: Isn't this now the same as line 21?

lwrubel: Thank you, fixed.

PR #935, opened 2024-10-24
FilesByMd5Service: in addition to confirming that a file exists before including it in the array #call returns, confirm that the file is of the expected size

jmartin-sul:
fixes https://github.com/sul-dlss/dor-services-app/issues/5191

assumes `size` metadata will always be present, but that seems like a safe bet?

### DIFF
--- a/app/services/versioned_files_service/cocina.rb
+++ b/app/services/versioned_files_service/cocina.rb
@@ -1,6 +1,8 @@
 class VersionedFilesService
   # Class for Cocina that doesn't rely on Dro::Models, since no guarantee that JSON will conform with latest.
   class Cocina
+    FileDetails = Struct.new('FileDetails', :md5, :filename, :filesize)
+
     # @param [String] druid the druid
     # @param [String] version the version
     # @return [Cocina] the Cocina object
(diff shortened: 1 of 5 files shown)


app/services/versioned_files_service/cocina.rb
```diff
@@ -24,12 +24,17 @@ def shelve_file_map
 
     # @return [Array<Hash<String, String>>] array of hashes with md5 as key and filename as value for shelved files
     # For example: [
-    #   { "5b79c8570b7ef582735f912aa24ce5f2" => "2542A.tiff" },
-    #   { "cd5ca5c4666cfd5ce0e9dc8c83461d7a" => "2542A.jp2" }
+    #   { "5b79c8570b7ef582735f912aa24ce5f2" => { "filename" => "2542A.tiff", "size" => 456 } },
```
justinlittman, line 27: I now regret the original data structure which has morphed into ^^. To avoid confusing our future selves, can we define a struct (or similar) to hold file details (with md5, filename, and size attributes) and then return an array of file details?

justinlittman: (Happy to discuss more / pair.)

justinlittman (changes requested):

app/services/versioned_files_service/cocina.rb
jmartin-sul: yup, that makes sense to me.  will give it a shot and discuss more or ask for pairing if i get stuck

jmartin-sul: ok, this is ready for another review.  i definitely prefer things with the struct suggestion, feels easier to read than the one element hash.  left the change as a separate commit so that it'd be easy to undo if the change i made wasn't what you had in mind.

i realized after i finished that commit that i didn't switch FilesByMd5Service.call from returning the md5 keyed hash list, but i guess since the return value gets used in a controller that other services presumably depend on, best to leave that structure as is?

jmartin-sul:
passed integration tests, merging

PR #971, opened 2025-03-06
Store and publish content types for collection members

thatbudakguy:
This implements storage and publishing of content type metadata
for purls, so that clients who query for collection members also
know the content type of each member.

See https://github.com/sul-dlss/purl/issues/84#issuecomment-2691156904
for discussion and a use case.


jcoyne:
@thatbudakguy which client would be querying purl-fetcher for the content type of each member?  Perhaps we should discuss the big picture here (at the access meeting) before proceeding with this.

jcoyne (approved):

PR #1007, opened 2025-08-04
Add an index on the purls.content_type column

jcoyne:


marlo-longley:
@jcoyne needs to be rebased

PR #1012, opened 2025-08-06
Analyze every purl to see if we need to migrate it or not

### LINKED ISSUE #1011: Create a script to find all druids that have not been migrated to versioned layout
Here is a legacy item: /stacks/fg/485/kx/8389

This should find all of the druids that are not migrated ( e.g. no /stacks/%{druid_tree}/%{druid}/versions/versions.json exists)

See /stacks/bb/198/rx/8176/bb198rx8176 as an example of a migrated version.

This is potentially millions of druids.

jcoyne:
fixes #1011

### DIFF
--- a/spec/models/version_migration_item_spec.rb
+++ b/spec/models/version_migration_item_spec.rb
@@ -11,4 +11,29 @@ RSpec.describe VersionMigrationItem do
       expect(described_class.last.status).to eq 'not_analyzed'
     end
   end
+
+  describe '.analyze_purls' do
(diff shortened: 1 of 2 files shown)


spec/models/version_migration_item_spec.rb
```diff
@@ -11,4 +11,30 @@
       expect(described_class.last.status).to eq 'not_analyzed'
     end
   end
+
+  describe '.analyze_purls' do
+    let(:stacks_pathname) { 'tmp/stacks' }
```
dnoneill, line 16: what is this for?

PR #1025, opened 2025-08-08
Benchmark creating Globus link

marlo-longley:


jcoyne:
We found that it was up to 1 minute on druids with 36,789 files.

PR #1031, opened 2025-08-13
Prevent duplicate DOI identifiers in public XML

mjgiarlo:
- **Fix flappy spec**
- **Prevent duplicate DOI identifiers in public XML**


spec/requests/v1/publish_collection_spec.rb
```diff
@@ -23,6 +23,10 @@
     end
 
     context 'when using the legacy endpoint' do
+      after do
+        FileUtils.rm_rf(Settings.filesystems.stacks_root)
+      end
```
mjgiarlo, line 28: This is unrelated to my change, but I got annoyed by this test failing when the suite was run in a certain order due to this test not cleaning up after itself.

jcoyne:
How/where is this causing a problem? Can you share an example?

mjgiarlo:
@jcoyne 💬 
> How/where is this causing a problem? Can you share an example?

Just finished writing it up: https://github.com/sul-dlss/purl-fetcher/pull/1031#issue-3320175973

PR #1055, opened 2025-09-29
Store compact cocina

### LINKED ISSUE #1022: Store compacted form of public Cocina
This is a proposal that maybe needs some further thought/discussion.

Currently, the public Cocina that we write contains many null or empty values – probably because the basic building block of Cocina, the `DescriptiveValue`, can be nested deeply and includes many generic attributes like `identifier` by default.

Here's an example of a date range (1932-2012) encoded in an event, taken from `druid:bb099mt5053`:

```json
{
        "structuredValue": [],
        "date": [
          {
            "structuredValue": [
              {
                "structuredValue": [],
                "parallelValue": [],
                "groupedValue": [],
                "value": "1932",
                "type": "start",
                "status": "primary",
                "identifier": [],
                "note": [],
                "appliesTo": []
              },
              {
                "structuredValue": [],
                "parallelValue": [],
                "groupedValue": [],
                "value": "2012",
                "type": "end",
                "identifier": [],
                "note": [],
                "appliesTo": []
              }
            ],
            "parallelValue": [],
            "groupedValue": [],
            "type": "creation",
            "encoding": { "code": "w3cdtf", "note": [] },
            "identifier": [],
            "qualifier": "approximate",
            "note": [],
            "appliesTo": []
          }
        ],
        "contributor": [],
        "location": [],
        "identifier": [],
        "note": [],
        "parallelEvent": []
      }
```

And here's that same value, but with empty keys and values removed, kind of like activesupport's `#compact`:

```json
{
        "date": [
          {
            "structuredValue": [
              {
                "value": "1932",
                "type": "start",
                "status": "primary",
              },
              {
                "value": "2012",
                "type": "end",
              }
            ],
            "type": "creation",
            "encoding": { "code": "w3cdtf" },
            "qualifier": "approximate",
          }
        ],
      }
```

It's worth considering whether we could perform this type of "deep compacting" automatically at the time we write the public Cocina, probably around here:
https://github.com/sul-dlss/purl-fetcher/blob/1774c01fedf5678953f1a046273a73104b249a8c/app/services/update_purl_metadata_service.rb#L26-L28

Some benefits:
- Less storage space used on disk
- Quicker transfer over the wire for apps that access the public cocina (e.g. at index time)
- Simpler code to parse the cocina (less null checks)
- Easier to read and debug

Basic testing indicates that space savings are highly variable; objects with many files but little descriptive metadata tend not to benefit as much as (smaller) objects with a lot of description.

|druid|existing json|compact json|savings|
|-|-|-|-|
|qr918wy2257|413kb|358kb|13%|
|vk217bh4910|190kb|92kb|51%|
|nz187ct8959|22kb|6.1kb|72%|
|bt553vr2845|13kb|3.2kb|75%|

This testing was done with [an existing algorithm](https://github.com/sul-dlss/cocina_display/blob/e281cc0cc6c02ca0ec633b44e790340a87be5615/lib/cocina_display/utils.rb#L51-L72) that's currently part of cocina_display; not used anywhere in production.

Related to/may alleviate #628.

jcoyne:
This will reduce storage size and parsing time

Fixes #1022

### DIFF
--- a/app/services/versioned_files_service/metadata.rb
+++ b/app/services/versioned_files_service/metadata.rb
@@ -12,7 +12,24 @@ class VersionedFilesService
     def write_cocina(version:, cocina:)
       FileUtils.mkdir_p(versions_path)
       cocina_path = cocina_path_for(version:)
-      cocina_path.write(cocina.to_json)
+      cocina_path.write(self.class.deep_compact_blank(cocina.to_h).to_json)
+    end
+
+    def self.deep_compact_blank(node)
+      return node unless node.is_a?(Hash)
+
+      node.each_with_object({}) do |(key, value), output|
+        case value
+        when Hash
+          nested = deep_compact_blank(value)
+          output[key] = nested unless nested.empty?
+        when Array
+          compacted_array = value.map { |v| deep_compact_blank(v) }.compact_blank
+          output[key] = compacted_array unless compacted_array.empty?
+        else
+          output[key] = value if value.present?
+        end
+      end
     end
 
     # Write the public xml file for a version.
--- a/spec/services/versioned_files_service_spec.rb
+++ b/spec/services/versioned_files_service_spec.rb
@@ -25,6 +25,7 @@ RSpec.describe VersionedFilesService do
   describe '#update' do
     let(:access_transfer_stage) { 'tmp/access-transfer-stage' }
     let(:version_metadata) { VersionedFilesService::VersionsManifest::VersionMetadata.new(version: 1, state: 'available', date: DateTime.now) }
+    let(:compact_cocina) { VersionedFilesService::Metadata.deep_compact_blank(cocina.to_h).to_json }
 
     before do
       FileUtils.mkdir_p(access_transfer_stage)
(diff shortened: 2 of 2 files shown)


app/services/versioned_files_service/metadata.rb
```diff
@@ -12,7 +12,32 @@ def initialize(paths:)
...
+          # Compact arrays by removing nils and empty hashes/arrays within them
+          compacted_array = value.filter_map do |item|
+            if item.is_a?(Hash)
+              deep_compact_hash(item)
+            elsif item.is_a?(Array)
+              # Recursively deep compact nested arrays if needed
+              item.filter_map { |sub_item| sub_item.is_a?(Hash) ? deep_compact_hash(sub_item) : sub_item }.reject(&:empty?)
+            else
+              item
+            end
```
thatbudakguy, line 34: Do we have to do these checks here, rather than just letting them happen when we recurse? The [cocina_display implementation](https://github.com/sul-dlss/cocina_display/blob/d0118d9e73a15640d07a9f889e0cc35ce55bd507/lib/cocina_display/utils.rb#L53-L77) ended up less verbose. Maybe I missed something, though?

thatbudakguy (approved):

PR #1058, opened 2025-09-30
Write virtual object memberships into the Cocina

### LINKED ISSUE #1057: (Virtual object issue) Constituent items should have a link back to their parent item
When an item is part of a virtual object, there should be a link on the purl back to the parent virtual object.

For example, if a map is:

- an item
- part of an atlas that is its own item

then the purl for the map should link back to the purl of the atlas.

This relationship is visible in SearchWorks ([link to example](https://searchworks.stanford.edu/view/rz617vr4473)):

Image

However, there is no longer a link back to the parent item on the [map's purl](https://purl.stanford.edu/rz617vr4473).

I think the problem is that the virtual object parent is only in the public XML and not in the public cocina. We need to get that data into the purl in a non-XML way. In the public XML, the parent is visible in two places:

Fedora RELS-EXT (vestigial trace of Fedora, probably not used in display):
```xml
<rdf:RDF xmlns:fedora="info:fedora/fedora-system:def/relations-external#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
    <rdf:Description rdf:about="info:fedora/druid:rz617vr4473">
      <fedora:isMemberOfCollection rdf:resource="info:fedora/druid:xh235dd9059"/>
      <fedora:isConstituentOf rdf:resource="info:fedora/druid:cp799hh4428"/>
    </rdf:Description>
```
and MODS `relatedItem`, which is probably where the display came from previously.
```xml
   <relatedItem type="host" displayLabel="Appears in">
      <titleInfo>
        <title>The Cyclers' Guide and Road Book of California Containing Map of California in relief with principal Roads, Seven Sectional Maps showing all available Roads for Cyclers from Chico to San Diego, and a Map of Golden Gate Park. 1896. Price, One Dollar. Compiled and Published by Geo. W. Blum, 330 Pine St., S.F. Edward Denny &amp; Co., Agents. (on verso) Copyrighted 1895 By Geo. W. Blum, San Francisco, Cal</title>
      </titleInfo>
      <location>
        <url>https://purl.stanford.edu/cp799hh4428</url>
      </location>
    </relatedItem>
```
In the old purl display logic, this "Appears in" entry was in every virtual object constituent purl.

jcoyne:
We currently do this for the xml, but not the cocina. When we switched purl to look at the cocina, we no longer drew this relationship on the page

Fixes #1057

### DIFF
--- a/app/services/publish/public_cocina_generator.rb
+++ b/app/services/publish/public_cocina_generator.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module Publish
+  # Adds the constituent (virtual object) relationships to the cocina object
+  class PublicCocinaGenerator
+    # @param [Cocina::Models::DRO, Cocina::Models::Collection] cocina
+    def initialize(cocina:)
+      @cocina = cocina
+    end
+
+    def self.generate(cocina:)
+      new(cocina: cocina).generate
+    end
+
+    def generate
+      related_resources = @cocina.description.relatedResource
+
+      augmented_description = @cocina.description.new(relatedResource: related_resources + constituent_resources)
+      @cocina.new(description: augmented_description)
+    end
+
+    private
+
+    def constituent_resources
+      VirtualObject.for(druid: @cocina.externalIdentifier).map do |virtual_object|
+        Cocina::Models::RelatedResource.new(title: [{ value: virtual_object[:title] }],
+                                            type: 'part of',
+                                            purl: purl_url(virtual_object[:id]),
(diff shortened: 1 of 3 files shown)


dnoneill (approved):
Do we still need https://github.com/sul-dlss/purl-fetcher/blob/main/spec/services/publish/public_desc_metadata_service_spec.rb? Should be migrate it to cocina?

jcoyne:
@dnoneill that is what Searchworks is currently using for indexing into Solr, so it can't be removed.

PR #1161, opened 2026-07-13
Update dependencies

sul-devops-team:


jcoyne (approved):

app/services/publish/public_xml_service.rb
```diff
@@ -99,5 +99,9 @@ def public_content_metadata
         )
       )
     end
+
+    def label
+      CocinaDisplay::CocinaRecord.new(public_cocina.to_h.deep_stringify_keys).display_title
```
justinlittman, line 104: `display_title` is the new `label` (which is deprecated).

app/services/to_mods/mods_writer.rb
```diff
@@ -26,8 +26,10 @@ def self.write(xml:, description:, druid:, id_generator: IdGenerator.new, identi
       Access.write(xml: xml, access: description.access,
                    purl: description.respond_to?(:purl) ? description.purl : nil)
       AdminMetadata.write(xml: xml, admin_metadata: description.adminMetadata)
-      RelatedResource.write(xml: xml, related_resources: description.relatedResource, druid: druid,
-                            id_generator: id_generator)
+      if description.respond_to?(:relatedResource)
```
justinlittman, line 29: `relatedResource` can no longer be nested, so test for presence of the method.

spec/mapping/mods/identifier_spec.rb
```diff
@@ -223,27 +223,6 @@
       end
     end
 
-    context 'when cocina is array with empty hash' do
```
justinlittman, line 226: These sorts of cases are now invalid.

PR #1162, opened 2026-07-14
Public label derived from title

lwrubel:
When writing public cocina to the filesystem, adds a label based on the title, using CocinaDisplay (added as a dependency). This sets up the public cocina to not be dependent on a label in an incoming cocina object. A public `label` is needed for digital bookplates and in any unknown situations where we are not already using a CocinaRecord for providing a title/label. 

Got some help from Claude, esp. with specs set-up.

Tested on stage by republishing/releasing an edited item, and the label in the public cocina is updated to match the edited title.

### DIFF
--- a/spec/services/versioned_files_service/update_action_spec.rb
+++ b/spec/services/versioned_files_service/update_action_spec.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe VersionedFilesService::UpdateAction do
+  let(:action) { described_class.new(version: 1, version_metadata:, cocina:, file_transfers: {}, object:) }
+
+  let(:druid) { 'druid:bc123df4567' }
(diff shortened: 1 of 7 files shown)


spec/services/versioned_files_service/update_action_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  describe '#call' do
+    it 'writes the label from the public cocina into cocina.json' do
+      action.call
+
+      written = JSON.parse(File.read("#{versions_path}/cocina.1.json"))
+      expect(written['label']).to eq 'Main title'
+    end
+
+    context "when the cocina object's own label differs from its title" do
+      let(:cocina) { build(:dro_with_metadata, id: druid).new(label: 'Some other label', access: { view: 'world', download: 'world' }) }
```
lwrubel, line 38: Will need to come back and update this when `label` is removed from the cocina-models schema.

jcoyne (approved):

PR #1164, opened 2026-07-16
Bump cocina-models

lwrubel:
The FromMods mapping of subjectAuthority changed in cocina-models with: https://github.com/sul-dlss/cocina-models/pull/1149/changes.

Should ToMods (now in purl-fetcher) also map this differently?  cc @arcadiafalcone


spec/mapping/mods/subject_cartographic_spec.rb
```diff
@@ -98,7 +98,7 @@
               type: 'map projection',
               uri: 'http://opengis.net/def/crs/EPSG/0/4326',
               source: {
-                code: 'EPSG'
+                value: 'EPSG'
```
jcoyne, line 101: Shouldn't this be `EPSG:4326`?  `EPSG` is like a namespace for a bunch of different projections. See https://spatialreference.org/ref/epsg/

jcoyne (approved):

jcoyne (approved):

PR #1167, opened 2026-07-22
Use a hash when reading stored public cocina

lwrubel:
Stored public cocina can be invalid cocina. Building a Cocina::Models object from the stored JSON can raise `Cocina::Models::ValidationError`s. This will certainly be the case when we remove the top-level `label` property from the cocina-models schema. Public cocina was being read for virtual object members when the parent is being republished. 

This changes some locations that used a cocina object to work with a hash instead. `CocinaDisplay` is used where possible. 

There are a few locations where a Cocina object is still being built, but not from cocina on stacks. 
* Argo POSTs to `ModsController` to get back MODS XML when the "Download descriptive metadata as XML" bulk action is used. That is only used for a current Cocina object which should be valid. 
* A publish request that goes through `Publish::PublicCocinaGenerator.generate(cocina:)` is also still using a cocina object. That could be looked at further in another update if needed.
 
Claude assisted with the analysis and code. 

Tested by republishing the canonical virtual object and also republishing a virtual object where I edited the stored cocina of a member to be invalid. Also tested with integration tests that do normal publishing.

### DIFF
--- a/app/models/cocina_data.rb
+++ b/app/models/cocina_data.rb
@@ -50,7 +50,7 @@ class CocinaData
   end
 
   def dro?
-    !collection?
+    Cocina::Models::DRO::TYPES.include?(data_hash['type'])
   end
 
   def catalog_record_ids(catalog)
--- a/app/models/cocina_data.rb
+++ b/app/models/cocina_data.rb
@@ -1,5 +1,5 @@
 class CocinaData
-  # @param [Hash] data_hash a parsed Cocina JSON document (Collection or DRO) with string keys
+  # @param [Hash{String => Object}] cocina_hash a hash parsed from a Cocina JSON document (either Collection or DRO)
   def initialize(data_hash)
     @data_hash = data_hash
   end
@@ -33,7 +33,7 @@ class CocinaData
 
   # @return [Array<String>] The collections the item is a member of
   def collections
-    dro? ? Array(data_hash.dig('structural', 'isMemberOf')) : []
+    Array(data_hash.dig('structural', 'isMemberOf'))
   end
 
   # @return [Array<String>] The constituent druids of this object (virtual object)
--- a/app/models/cocina_data.rb
+++ b/app/models/cocina_data.rb
@@ -1,14 +1,14 @@
 class CocinaData
-  # @param [Cocina::Models::Collection, Cocina::Models::DRO]
-  def initialize(cocina_object)
-    @cocina_object = cocina_object
+  # @param [Hash] data_hash a parsed Cocina JSON document (Collection or DRO) with string keys
+  def initialize(data_hash)
+    @data_hash = data_hash
   end
 
-  attr_reader :cocina_object
+  attr_reader :data_hash
 
   # @return [String] The druid in the form of druid:pid
   def canonical_druid
-    cocina_object.externalIdentifier
+    data_hash['externalIdentifier']
   end
 
   # @return [String] The catkey. An empty string is returned if there is no catkey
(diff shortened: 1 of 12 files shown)


app/services/thumbnail_service.rb
```diff
@@ -5,27 +5,27 @@ class ThumbnailService
   # allow the mimetype attribute to be lower or camelcase when searching to make it more robust
   MIME_TYPE = 'image/jp2'
 
-  # @param [Cocina::Model::DRO] object
+  # @param [Cocina::Models::DRO, Hash] object
```
lwrubel, line 8: Could be a Hash if it's a virtual object member and the `thumb` call is returning a Hash via CocinaObjectStore (L29). A Cocina object would the param if this is the original publish request.

app/services/cocina/to_xml/content_metadata_generator.rb
```diff
@@ -59,10 +59,10 @@ def add_members_data
 
         index = 0
         members.each do |external_druid|
-          cocina_object = cocina_object_store.find(external_druid)
-          Array(cocina_object.structural.contains).each do |cocina_fileset|
+          cocina_hash = cocina_object_store.find(external_druid)
+          label = CocinaDisplay::CocinaRecord.new(cocina_hash).display_title
```
lwrubel, line 63: Calculate top-level label once.

app/services/cocina_object_store.rb
```diff
@@ -4,7 +4,6 @@ def self.find(druid)
 
     raise "No cocina.json found for #{druid} in stacks or purl paths" unless stacks_path.exist?
 
-    data_hash = JSON.parse(File.read(stacks_path))
-    Cocina::Models.build(data_hash)
```
lwrubel, line 8: Main change here: no longer build a Cocina object from stored public cocina.

app/models/cocina_data.rb
```diff
@@ -18,37 +18,42 @@ def catkey
...
   def content_type
-    Cocina::ToXml::ContentType.map(cocina_object.type) if cocina_object.dro?
+    Cocina::ToXml::ContentType.map(data_hash['type']) if dro?
   end
 
   # @return [Array<String>] The collections the item is a member of
   def collections
-    # see https://github.com/sul-dlss/dor-services-app/blob/f51bbcea710b7612f251a3922c5164ec69ba39aa/app/services/published_relationships_filter.rb#L31
-    cocina_object.dro? ? cocina_object.structural.isMemberOf : []
+    dro? ? Array(data_hash.dig('structural', 'isMemberOf')) : []
```
mjgiarlo: I believe `Hash#dig` is safe to use on hashes that lack the specified dig args, so I wonder if we could replace this with:

```suggestion
    Array(data_hash.dig('structural', 'isMemberOf'))
```

See this little experiment:

```
irb(main):001> h = {}
=> {}
irb(main):002> h.dig('structural', 'isMemberOf')
=> nil
irb(main):003> Array(h.dig('structural', 'isMemberOf'))
=> []
```

app/models/cocina_data.rb
```diff
@@ -1,14 +1,14 @@
 class CocinaData
-  # @param [Cocina::Models::Collection, Cocina::Models::DRO]
-  def initialize(cocina_object)
-    @cocina_object = cocina_object
+  # @param [Hash] data_hash a parsed Cocina JSON document (Collection or DRO) with string keys
+  def initialize(data_hash)
+    @data_hash = data_hash
   end
```
mjgiarlo: Some non-blocking thoughts:

```suggestion
  # @param [Hash{String => Object}] cocina_hash a hash parsed from a Cocina JSON document (either Collection or DRO)
  def initialize(cocina_hash)
    @cocina_hash = cocina_hash
end
```

This makes two changes:

1. Convert the comment text saying the hash has string keys into YARD documentation
2. Renames `data_hash` to `cocina_hash` to be more intentional-revealing throughout the model, and aligning the naming with what's in the `Purl` model.

Again, non-blocking! Throw this away if you're not a fan.

app/services/thumbnail_service.rb
```diff
@@ -5,27 +5,27 @@ class ThumbnailService
...
   def initialize(object)
-    @object = object
+    data_hash = object.is_a?(Hash) ? object : object.to_h.deep_stringify_keys
+    @record = CocinaDisplay::CocinaRecord.new(data_hash)
   end
```
mjgiarlo, line 12: Another non-blocking simplification idea:

```suggestion
  # @param [#to_h] object
  def initialize(object)
    @record = CocinaDisplay::CocinaRecord.new(object.to_h.deep_stringify_keys)
  end
```

i.e., if the object is already a hash, it should be safe to call `#to_h` on it and (re-)stringify the keys.

mjgiarlo: And the YARD change says "this class can deal with any object that responds to `to_h`"

app/services/thumbnail_service.rb
```diff
@@ -5,27 +5,27 @@ class ThumbnailService
...
+    record.filesets.each do |file_set|
+      file_set.files.each do |file|
+        next unless file.mime_type == MIME_TYPE
 
-    object.structural.contains.each do |file_set|
-      file_set.structural.contains.each do |file|
-        next unless file.hasMimeType == MIME_TYPE
-
-        return "#{object.externalIdentifier.delete_prefix('druid:')}/#{file.filename}"
+        return "#{record.bare_druid}/#{file.filename}"
```
mjgiarlo, line 22: So this block short-circuits this method on the first file matching `MIME_TYPE`?

mjgiarlo (commented):
Approve. Non-blocking suggestions only.

mjgiarlo (approved):
Oops, meant to approve in my last review.

app/models/cocina_data.rb
lwrubel: Much more concise, thanks!

app/services/thumbnail_service.rb
lwrubel: I think I'll leave this one as-is for now, just to help in highlighting where Cocina objects are being passed around.

app/services/thumbnail_service.rb
lwrubel: The code is looking for the jp2 thumbnail. I think the expectation is that there is only one jp2 (and the first one. That shortcircuit logic is unchanged from before.

